PR #3, opened 2015-01-30
Feature/aku 24

draperd:
This pull request addresses https://issues.alfresco.com/jira/browse/AKU-24 which observes that it is possible exploit an XSS vulnerability when creating select form controls with scripting as values. This is because the widgets are created as fragments programmatically and then added to the document. The changes in this PR update the behaviour of all form controls so that value setting doesn't occur until the widget is added to the document. This ensures that no malicious values will be executed. In the process of closing the vulnerability it was necessary to make changes to other widgets because unit tests failed. Many unit tests have been re-written to the preferred style as an aid to debugging the test failures. There is still an outstanding issue on the drag-and-drop creation widgets, but the test has been quarantined as this is non-Alfresco production code. It will be investigated and addressed in a forthcoming sprint - the most important thing was to close the vulnerability for widgets that are used in production.


aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -214,7 +215,7 @@ define(["dojo/_base/declare",
        */
       publishFormValidity: function alfresco_forms_Form__publishFormValidity() {
          this.alfPublish("ALF_FORM_VALIDITY", {
-            valid: (this.invalidFormControls.length === 0),
+            valid: this.invalidFormControls.length === 0,
```
MartinDoyleUK: `valid: !this.invalidFormControls.length` would also work here

aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -461,10 +462,11 @@ define(["dojo/_base/declare",
       createButtons: function alfresco_forms_Form__createButtons() {
          if (this.showOkButton === true)
          {
+            
             this.okButton = new AlfButton({
                pubSubScope: this.pubSubScope,
                label: this.message(this.okButtonLabel),
-               additionalCssClasses: "confirmationButton " + ((this.okButtonClass != null) ? this.okButtonClass : ""),
+               additionalCssClasses: "confirmationButton " + (this.okButtonClass ? this.okButtonClass : ""),
```
MartinDoyleUK: `this.okButtonClass || ""` is shortest form, if optimising

aikau/src/main/resources/alfresco/logging/SubscriptionLog.js
```diff
@@ -217,7 +217,9 @@ define(["dojo/_base/declare",
        * @param {element} domNode The element to add the string to
        */
       addStringToLog: function alfresco_testing_Subscription__addStringToLog(s, domNode) {
-         domNode.innerHTML = s;
+         domConstruct.empty(domNode);
+         domNode.appendChild(document.createTextNode(s));
+         // domNode.innerHTML = s;
```
MartinDoyleUK: This line not needed?

MartinDoyleUK:
Reviewed and merged with advisory comments.

PR #59, opened 2015-02-23
Added additionalCssClasses support to TextArea and default CSS requirements.

kevinroast:
This ensures default CSS (for border etc.) of TextArea is same as other HTML form controls such as TextBox and that additionalCssClasses can be used.


kevinroast:
Missed a file, git skillz #fail.

PR #90, opened 2015-03-09
AKU-30: Indicators don't support action handling

MartinDoyleUK:
This addresses issue https://issues.alfresco.com/jira/browse/AKU-30, which is about indicators not supporting actions. It also updates the .gitignore and .jsbeautifyrc files and adds an (ES5) shim.


aikau/src/main/resources/alfresco/core/Page.js
```diff
@@ -227,7 +228,7 @@ define(["alfresco/core/ProcessWidgets",
        */
       _registerProcessedService: function alfresco_core_Core___registerProcessedService(service, index) {
          this._processedServiceCountdown--;
-         if (index == null || isNaN(index))
+         if (!index || isNaN(index))
```
draperd: What happens when index is 0 here do we think?

aikau/src/main/resources/alfresco/core/Page.js
```diff
@@ -192,7 +193,7 @@ define(["alfresco/core/ProcessWidgets",
             {
                // If there is a callback then call it with any provided scope (but default to the
                // "this" as the scope if one isn't provided).
-               callback.call((callbackScope != null ? callbackScope : _this), service, index);
+               callback.call((callbackScope || this), service, index);
```
draperd: I think "this" needs to be "_this"

aikau/src/main/resources/alfresco/core/Page.js
```diff
@@ -227,7 +228,7 @@ define(["alfresco/core/ProcessWidgets",
        */
       _registerProcessedService: function alfresco_core_Core___registerProcessedService(service, index) {
          this._processedServiceCountdown--;
-         if (index == null || isNaN(index))
+         if (index !== 0 && (!index || isNaN(index)))
```
draperd: I think this is still not quite right, we still want to push the service if it is at index 0

PR #111, opened 2015-03-13
AKU-115, AKU-116: Accessibility actions for DND

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-115 and https://issues.alfresco.com/jira/browse/AKU-116 and adds keyboard action handling for adding new items from a drag and drop palette into a drag and drop target and then allows those items to be re-ordered, edited and deleted. This includes an updated unit test (with one test temporarily commented out because of failure on Firefox which is hard to debug until I've resolved local selenium version and Firefox version issues)


aikau/src/main/resources/alfresco/dnd/DragAndDropItem.js
```diff
@@ -0,0 +1,83 @@
...
+       * the currently selected item in a [DragAndDropItems]{@link module:alfresco/dnd/DragAndDropItems}
+       * widget.
+       * 
+       * @instance
+       * @param {object} evt The keypress event
+       */
+      onKeyPress: function alfresco_dnd_DragAndDropTarget__onKeyPress(evt) {
+         if (evt.charOrCode === keys.ENTER)
+         {
+            evt && Event.stop(evt);
```
MartinDoyleUK: Must already have an event here, but could potentially break on the line above if no event is specified. Maybe move the evt && to the condition?

PR #121, opened 2015-03-18
AKU-174: Added a strict mode to BUILD payload generation

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-174. It adds the option to make enforce strict existence of data (by setting the "alfStrict" attribute).

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/_PublishPayloadMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_PublishPayloadMixin.js
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2005-2014 Alfresco Software Limited.
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
  *
  * This file is part of Alfresco
  *
@@ -204,6 +204,7 @@ define(["dojo/_base/declare",
             {
                var type = value.alfType;
                var property = value.alfProperty;
+               var strict = value.alfStrict;
 
                if (type === "item" && currentItem)
                {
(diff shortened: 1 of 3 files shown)


draperd:
Please wait for comments on suggested update before pulling.

kevinroast:
>The reason for not always failing is that I can see cases where we want to allow missing data...
What cases? When would "undefined" be acceptable return value for this method?

>I can always flip the default for it to be strict mode (only 1.0.9 will have not failed on missing data).
It's basically useless unless script is the default. Rather than "strict" boolean - fix up the cases where undefined is allowed through - which are probably bugs.

draperd:
Consider the situation we currently have in the Document Library where we set a location either by a combination of "site" and "container" OR "rootNode". We may want to trigger a BUILT payload off this data, so "site" and "container" not being present might be perfectly acceptable in the resulting payload. 

The point here is that we never know what people might want to do, and I'd like to keep options open rather than mandating something has to fail if not essential. There's nothing wrong with generating a payload that has null/undefined attributes set on it.

spotnape:
The example we had in CMM was a model created with optional fields left empty and then returned absent from the API during a subsequent GET. When using the payload build function for the edit dialogue, the missing fields caused an exception and end of processing rather than just being created on the payload as empty values.

kevinroast:
The framework should not blow up futher up the stack (which is what is happening) when undefined is returned from this method then. The problem is that the calling code has no idea this has happened - it just stops processing - and there is no useful error reported.

Instead then, can you handle the undefined better in the BUILD case for forms? i.e. the form processing check for undefined in the excepted payload and throw the error from there? The problem is the developer has no chance to deal with this error because it all happens inside Aikau.

draperd:
OK, that certainly seems like a reasonable request (to prevent forms blowing up without error/warning). The BUILD payload is not just used for forms so I'd prefer to keep it flexible, but I am happy to make "strict" the default behaviour (and rename the configuration attribute). I'd like to further understand the form problem and maybe address it in a different issue.

PR #126, opened 2015-03-19
Dialog display updates

draperd:
Ensure that footer is always displayed on the dialog and overflow is handled appropriately. This undoes the change made in #112 (which prevented the tab key from getting to the dialog buttons) but addresses the issue of making the footer always visible through CSS.


draperd:
Wait for review by Kev (who made #112) before pulling

kevinroast:
I'm ok if you can assure me the scrolling of the body works correctly, then go for it.

draperd:
Unfortunately IE8 and IE9 don't support calc at all/very well (http://caniuse.com/#feat=calc) I'm considering that we set specific CSS rules for those browsers so that the footer isn't always visible - not ideal, but these browsers should be increasingly be a minority

draperd:
Re-basing forced me to close and I've re-opened here: https://github.com/Alfresco/Aikau/pull/128

PR #127, opened 2015-03-20
Furrther DND updates

draperd:
These updates to the DND capabilities add in the ability to support form specific model creation (and provides a form specific modelling library file that can be included in pages). It also adds the ability for nested items to inherit configuration from the parent they have been dropped into. Unit tests have been added to verify all the changes. Other fixes are included relating to PageService and FormControls.


aikau/src/main/resources/alfresco/dnd/DragAndDropFormControlTarget.js
```diff
@@ -0,0 +1,95 @@
...
+               }
+               else
+               {
+                  this.alfLog("log", "Value and label properties are not set on form control yet so cannot be added as an option", this, value);
+               }
+            }
+         }, this);
+
+         // Update all the other dropped widgets so that they have a reference to them...
+         array.forEach(nodes, function(node) {
```
MartinDoyleUK: Don't need two loops here. Because the array is an object, its reference can be added to each widget, and each widget will be updated when the array is pushed into. This is a performance overhead (especially with large numbers of nodes), but is less legible than having a separate loop, however a simple comment may suffice to explain what's going on.

draperd:
I've made the change as suggested, as well as a correction to the unit test (which I fortunately diligently ran to verify the change!)

PR #150, opened 2015-04-01
Refactor token replacement so it works properly with missing tokens and ...

davidcognite:
...doesn't have duplicated code. Extends tests to cover this.


davidcognite:
the unified diff doesn't show the ObjectProcessingMixin changes very well. I've extracted the token replacement code into a utility widget that is used by both processInstanceTokens and processCurrentItemTokens and - crucially - this util method returns the original value if it wasn't replaced. 

Previously behaviour was that for a single missing token it would return undefined (even following my fix yesterday), but for multiple tokens (i.e. if lang.replace was used) it worked as expected.

PR #154, opened 2015-04-02
AKU-192: Added support for list filtering

draperd:
This addresses https://issues.alfresco.com/jira/browse/AKU-192

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfList.js
@@ -277,6 +277,100 @@ define(["dojo/_base/declare",
             // this.processWidgets(JSON.parse(JSON.stringify(this.widgets)));
             this.processWidgets(this.widgets);
          }
+
+         if (this.filteringTopics)
+         {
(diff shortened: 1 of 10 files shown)


aikau/src/main/resources/alfresco/lists/AlfList.js
```diff
@@ -531,7 +628,7 @@ define(["dojo/_base/declare",
        * @return {boolean} Either true or false depending upon the validity of the supplied configuration.
        */
       isValidViewSelectionConfig: function alfresco_lists_AlfList__isValidViewSelectionConfig(viewSelectionConfig) {
-         return (viewSelectionConfig.label != null && viewSelectionConfig.label !== "");
+         return (viewSelectionConfig.label && viewSelectionConfig.label !== "");
```
MartinDoyleUK: Can't be truthy and an empty string ...

draperd: But I want to return a boolean, not the label value... hence the && statement.

PR #160, opened 2015-04-10
AKU-191: Creation of a DND items list view

draperd:
This addresses https://issues.alfresco.com/jira/browse/AKU-191.


aikau/src/main/resources/alfresco/lists/views/layouts/_MultiItemRendererMixin.js
```diff
@@ -403,16 +403,16 @@ define(["dojo/_base/declare",
        */
       createWidget: function alfresco_lists_views_layout___MultiItemRendererMixin__createWidget(config, domNode, callback, callbackArgs) {
          // Only create a widget if there is data to create it with
-         if (config == null || config.config == null)
+         if (!config || !config.config)
          {
             config.config = {};
```
MartinDoyleUK: This will error if config is null

PR #162, opened 2015-04-13
AKU-178: Multi-select input box

MartinDoyleUK:
This addresses issue https://issues.alfresco.com/jira/browse/AKU-178 which requests the creation of an input box which allows multiple choices to be selected from a dropdown within the control.


aikau/src/main/resources/alfresco/forms/controls/MultiSelect.js
```diff
@@ -0,0 +1,1230 @@
...
+          * @type {object[]}
+          * @default [{i18nFile: "./i18n/MultiSelect.properties"}]
+          */
+         i18nRequirements: [{
+            i18nFile: "./i18n/MultiSelect.properties"
+         }],
+
+         /**
+          * The localised loading message used by the template
+          *
```
draperd: Need @instance here (and on all the other attributes too, otherwise it gets displayed as a global attribute in the final JSDoc

aikau/src/main/resources/alfresco/forms/controls/MultiSelect.js
```diff
@@ -0,0 +1,1230 @@
...
+
+            // Set the new value
+            this._changeAttrValue("value", chosenItems);
+         },
+
+         /**
+          * Add the specified result item to the choices
+          *
+          * @instance
+          * @protected
```
draperd: We need to be a bit careful about how we use JSDoc attributes like this as we're looking to somewhat abuse them to define what is and isn't safe from an extensibility point of view, this is still up for discussion so for the time being I'd like to avoid us using them to prevent possible confusion in the future

aikau/src/main/resources/alfresco/forms/controls/MultiSelectInput.js
```diff
@@ -0,0 +1,74 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @module alfresco/forms/controls/MultiSelectInput
```
draperd: It would be nice to get into the habit of adding in some meaningful JSDoc for widgets that we might expect developers to use, I'd quite like to see us get into the habit of using the @example attribute where there might be configuration that is useful to know. In the case of the form controls, most things should be common so it might not be too important to provide an example - but we should at least aim to provide a description of what the form control is.

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
+.alfresco-forms-controls-MultiSelect {
+   border: 1px solid #666;
```
draperd: Although they're not fully formed, we should be looking at re-using the LESS variables that are known and looking to abstract new ones where applicable. There is border variables, it may or may not be appropriate to use here, but there are no LESS variables referenced at all here.

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
...
+      &__content {
+         color: #333;
+         line-height: 22px;
+      }
+      &__close-button {
+         color: #aaa;
+         cursor: pointer;
+         display: block;
+         font-family: @bold-font;
+         font-size: 16px;
```
draperd: Should we be using the font size variable?

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
...
+         font-weight: 700;
```
draperd: Not sure if we want to use weight, here... if looking to make something bolder then we should use the bold font variable (which is a different font). We have tried to avoid setting weight at all.

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
...
+         &:hover {
+            color: #666;
+         }
+      }
+      &--selected {
+         background: #06a linear-gradient(to bottom, #08c, #06a);
+         border-color: #039;
+         .alfresco-forms-controls-MultiSelect__choice {
+            &__content {
+               color: #fff;
```
draperd: Possible candidate for LESS variable, is this just a background colour?

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
...
+               color: #cef;
+               &:hover {
+                  color: #fff;
+               }
+            }
+         }
+      }
+   }
+   &__off-screen-search {
+      font-size: 14px;
```
draperd: LESS variable

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
...
+      max-width: 100%;
+      padding: 0;
+      width: 50px;
+      &:focus {
+         outline: 0;
+      }
+   }
+   &__results {
+      background: #fff;
+      box-shadow: 2px 2px 3px rgba(0, 0, 0, .1);
```
draperd: I think we have a common box shadow LESS variable as well

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -0,0 +1,174 @@
...
+         text-overflow: ellipsis;
+         white-space: nowrap;
+      }
+   }
+   &__result-loading, &__result-error, &__result-no-results {
+      border-width: 1px;
+      display: none;
+   }
+   &__result-no-results {
+      color: #999;
```
draperd: Font colour is almost something we're going to want to be themable via LESS variables

aikau/src/main/resources/alfresco/forms/controls/MultiSelect.js
```diff
@@ -0,0 +1,1230 @@
...
+          */
+         _updateItemsFromStore: function alfresco_forms_controls_MultiSelect___updateItemsFromStore() {
+
+            // Make sure we have some that need updating
+            if (!this._itemsToUpdateFromStore.length) {
+               return;
+            }
+
+            // Setup handlers
+            var successHandler = lang.hitch(this, function(responseItems) {
```
draperd: Is there a possibility that someone (us or a 3rd party) might want to override the code in this closure? If so, we should probably move it out to a widget level function.

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/forms/controls/MultiSelectInput.get.desc.xml
```diff
@@ -0,0 +1,6 @@
+<webscript>
+  <shortname>MultiSelectInput Test</shortname>
```
draperd: We need to make these names and descriptions more useful as they're now displayed in the main unit test page, e.g. "MultiSelectInputFormControl" as the shortName, and something along the lines of "Shows an example of the alfresco/forms/controls/MultiSelectInput form control using mocked tag data"

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/forms/controls/MultiSelectInput.get.html.ftl
```diff
@@ -0,0 +1 @@
+<@processJsonModel group="share"/>
```
draperd: Keep meaning to say, but we don't need the group="share" anymore, that's a hangover from when the code was part of Share - you don't necessarily need to remove it (there are lots of other examples where this is still present) but be aware that it's no longer necessary)

draperd:
I've made a bunch of comments on the code, I can't see any major issues with this. I've pulled the branch and tried it out and it seems really good - nice work... it looks like there's no support yet for the create items that don't exist (which is fine - we can address in another issue) - I just wanted to confirm that was the case.

PR #174, opened 2015-04-20
AKU-227: CommentsList improvements

MartinDoyleUK:
Addresses most of the issues in [AKU-227](https://issues.alfresco.com/jira/browse/AKU-227). Some of the issues will be moved into separate stories, as they are more appropriately dealt with separately.

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/VerticalReveal.js
+++ b/aikau/src/main/resources/alfresco/layout/VerticalReveal.js
@@ -18,7 +18,18 @@
  */
 
 /**
- * This is a BETA quality widget and not guaranteed for production use.
+ * This layout widget allows for the dynamic revealing and hiding of content.
+ *
+ * @example <caption>Sample configuration</caption>
+ * {
+ *   name: "alfresco/layout/VerticalReveal",
+ *   config: {
+ *     subscriptionTopic: "ALF_REVEAL_LOGO",
+ *     widgets: [{
+ *       name: "alfresco/logo/Logo"
+ *     }]
+ *   }
+ * }
  *
  * @module alfresco/layout/VerticalReveal
  * @extends module:alfresco/core/ProcessWidgets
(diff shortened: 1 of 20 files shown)


aikau/src/main/resources/alfresco/layout/VerticalReveal.js
```diff
@@ -83,61 +92,94 @@ define(["alfresco/core/ProcessWidgets",
...
-            this.processWidgets(this.widgets, this.containerNode);
-            this._processedWidgetsToReveal = true;
-         }
+      /**
+       * Handle requests to toggle the contents visibility
+       *
+       * @instance
+       */
+      _onDisplayToggle: function alfresco_layout_VerticalReveal___onDisplayToggle() {
+         this._processWidgets().then(lang.hitch(this, function() {
```
draperd: We should avoid the use of the closure here, this is likely to be a prime example of something that might want to be overridden by an extending module.

aikau/src/main/resources/alfresco/layout/VerticalReveal.js
```diff
@@ -83,61 +92,94 @@ define(["alfresco/core/ProcessWidgets",
...
-            this.processWidgets(this.widgets, this.containerNode);
-            this._processedWidgetsToReveal = true;
-         }
-         var maxHeight = domStyle.get(this.contentNode, "maxHeight"),
-            isExpanded = !isNaN(maxHeight) && maxHeight > 0;
-         if (isExpanded) {
-            domStyle.set(this.contentNode, "maxHeight", 0);
-         } else {
-            domStyle.set(this.contentNode, "maxHeight", this.contentNode.scrollHeight + "px");
+      _processWidgets: function alfresco_layout_VerticalReveal___processWidgets() {
```
draperd: I'm not really sure what the benefit this approach is beyond simply including a promise. The code could surely be simpler if just calling processWidgets the first time and then the flag being set in allWidgetsProcessed. I'm not sure where the additional value is, especially as this then differs from the pattern used everywhere else.

aikau/src/main/resources/alfresco/renderers/EditableComment.js
```diff
@@ -250,6 +254,25 @@ define(["dojo/_base/declare",
       okButtonLabel: "Save",
```
draperd: This needs to be localized

aikau/src/main/resources/alfresco/layout/VerticalReveal.js
```diff
@@ -18,35 +18,36 @@
  */
 
 /**
- * This is a BETA quality widget and not guaranteed for production use. 
+ * This is a BETA quality widget and not guaranteed for production use.
```
draperd: We need better JSDoc here - I appreciate I should have done it the first time around, but we need to update the JSDoc on modules each time we edit them.

aikau/src/main/resources/alfresco/layout/VerticalReveal.js
MartinDoyleUK: If the toggle handler kicks off an asynchronous process to load the widgets that the allWidgetsProcessed method then finishes, how would we know to do the toggle after this flag has been set? Doing a looped setTimeout to check for the flag being set seems wrong, and what promises are intended to solve?
Are there any examples I could copy, where a similar situation has occurred, as I may well be not understanding how best to do this according to existing conventions.

PR #176, opened 2015-04-21
AKU-233: Thumbnail refactor and associated updates

draperd:
This PR initially addresses https://issues.alfresco.com/jira/browse/AKU-233 but has been used to go through the thumbnail renderer and all associated modules that it works with.

Along the way I've fixed up any JSHint issues that I've encountered. I've removed the legacy PDF.js plugin (that was uses a prototype hack) and swapped in the new Aikau version as the default. I've also had to pull in the 3rd party lightbox.js file from Share and although I've made some updates, it's still in quite a state - although it is functional, I've raised https://issues.alfresco.com/jira/browse/AKU-241 to address this 3rd party lib so we shouldn't dwell too much on the generally woeful JS that we've inherited.

I've made the SearchThumbnail use a common approach for showing previews and have added support for fetching node data so that any node can be previewed just by providing a nodeRef.

I've also updated the unit test so that it's actually doing something vaguely useful now.


aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -115,16 +159,26 @@ define(["dojo/_base/declare",
...
                this.thumbnailUrl = this.generateRenditionSpecificThumbnailUrl(nodeRef.uri);
-               if (this.thumbnailUrl === null)
+               if (!this.thumbnailUrl)
                {
                   this.thumbnailUrl = AlfConstants.PROXY_URI + "api/node/" + nodeRef.uri + "/content/thumbnails/" + this.renditionName + "/?c=queue&ph=true&lastModified=" + this.currentItem.modifiedOn;
                }
             }
+            else if (nodeRef && this.assumeRendition)
+            {
+               //
```
davidcognite: What's with the empty comment line?

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -115,16 +159,26 @@ define(["dojo/_base/declare",
             else if (this.currentItem.type === "document" && nodeRef.uri)
             {
                this.thumbnailUrl = this.generateRenditionSpecificThumbnailUrl(nodeRef.uri);
-               if (this.thumbnailUrl === null)
+               if (!this.thumbnailUrl)
                {
                   this.thumbnailUrl = AlfConstants.PROXY_URI + "api/node/" + nodeRef.uri + "/content/thumbnails/" + this.renditionName + "/?c=queue&ph=true&lastModified=" + this.currentItem.modifiedOn;
```
davidcognite: Should the thumbnail URL be extracted into a tokenised property?

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
draperd: A mistake :)

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
draperd: Yes, I've considered that as a future enhancement

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -275,89 +367,65 @@ define(["dojo/_base/declare",
...
-                        publishPayload: {
-                           nodeRef: this.currentItem.nodeRef
-                        }
-                     }
-                  ]
-               };
+               this.nodePromise = new Deferred();
+               this.onLoadNode(nodeRef);
             }
+            when(this.nodePromise, lang.hitch(this, this.onNodePromiseResolved, nodeRef));
```
davidcognite: Should this be an event to allow easier extensions?

draperd: Extensions can be done by extending onNodePromiseResolved if necessary

PR #177, opened 2015-04-21
AKU-245: Localize the CrudService messages

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-245 to ensure that all messages generated by the CrudService get localized. I've not added a unit test for this as it seems superfluous and is likely to be covered by other unit tests (e.g. editing/deleting comments, etc)

### DIFF
--- a/aikau/src/main/resources/alfresco/services/CrudService.js
+++ b/aikau/src/main/resources/alfresco/services/CrudService.js
@@ -82,7 +82,6 @@ define(["dojo/_base/declare",
             message = "crudservice.generic.success.message";
          }
 
-         // TODO: Need a context sensitive, localized message...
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/services/CrudService.js
```diff
@@ -79,12 +79,12 @@ define(["dojo/_base/declare",
 
          var message = lang.getObject("successMessage", false, originalRequestConfig);
          if (!message) {
-            message = this.message("crudservice.generic.success.message");
+            message = "crudservice.generic.success.message";
          }
 
          // TODO: Need a context sensitive, localized message...
```
davidcognite: This todo can be removed now, can't it?

PR #185, opened 2015-04-22
AKU-232: ActionService update and test

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-232. I'm continuing to move actions out of the ActionService and into more appropriate services so that they can be easily called outside of the Document Library. I've made the fix originally required for the issue and added a unit test that was missing for the delete action (and a few JShint fixes along the way).

### DIFF
--- a/aikau/src/main/resources/alfresco/services/ContentService.js
+++ b/aikau/src/main/resources/alfresco/services/ContentService.js
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2005-2014 Alfresco Software Limited.
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
  *
  * This file is part of Alfresco
  *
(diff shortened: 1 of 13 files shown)


aikau/src/main/resources/alfresco/services/ContentService.js
```diff
@@ -73,13 +74,12 @@ define(["dojo/_base/declare",
...
-
-         if (payload.alf_destination == null || payload.alf_destination === "")
+         if (!payload.alf_destination)
          {
             payload.alf_destination = this._currentNode.parent.nodeRef;
          }
          var type = null;
-         if (payload.type == null)
+         if (!payload.type)
          {
```
MartinDoyleUK: Could be written as `var type = payload.type || "cm%3acontent";`

PR #204, opened 2015-04-30
AKU-251: Update PreferenceService for favourites

draperd:
This PR addresses AKU-251 (and in part AKU-254) to support the adding and removal of favourites on an Aikau client that is independent of Share. Existing YUI/Share code dependencies have been removed and ported into the Aikau code base. The unit tests have been updated to include tests for both the PreferenceService and RatingsService and the doclib.lib.js has been updated to make it easier to get the list of services required for using the Document Library.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/PreferenceService.js
+++ b/aikau/src/main/resources/alfresco/services/PreferenceService.js
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2005-2013 Alfresco Software Limited.
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
  *
  * This file is part of Alfresco
  *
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/services/PreferenceService.js
```diff
@@ -98,11 +88,36 @@ define(["dojo/_base/declare",
...
+            {
+               property = properties[i];
+               obj[property] = {};
+               obj = obj[property];
+            }
+            obj[properties[i]] = value !== undefined ? value : null;
+         }
+         return object;
+      },
+
```
MartinDoyleUK: Why are we not using dojo/_base/lang.getObject() here?

aikau/src/test/resources/testApp/js/aikau/testing/mockservices/SocialMockXhr.js
```diff
@@ -0,0 +1,66 @@
...
+   
+   return declare([MockXhr], {
+
+      /**
+       * This sets up the fake server with all the responses it should provide.
+       *
+       * @instance
+       */
+      setupServer: function alfresco_testing_SitesPaginationMockXhr__setupServer() {
+         try
```
MartinDoyleUK: Wrong scoped method name

aikau/src/main/resources/alfresco/services/PreferenceService.js
draperd: Because I took the code verbatim from alfresco.js out of the Share project. Also, I'm not sure that getObject does what we want here... we want to create an object from a string, not inspect an existing object with a string

MartinDoyleUK: I think if you pass true as the second parameter, to [lang.getObject()](http://dojotoolkit.org/reference-guide/1.10/dojo/_base/lang.html#getobject) then it should create it?

PR #208, opened 2015-04-30
AKU-254: Ensure selection click event is stopped

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-257

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/views/AlfGalleryView.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/views/AlfGalleryView.js
@@ -31,9 +31,8 @@ define(["dojo/_base/declare",
         "dojo/text!./templates/AlfGalleryView.html",
         "alfresco/lists/views/layouts/Grid",
         "alfresco/documentlibrary/AlfGalleryViewSlider",
-        "dojo/_base/lang",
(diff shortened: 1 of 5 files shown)


aikau/src/main/resources/alfresco/documentlibrary/views/AlfGalleryView.js
```diff
@@ -97,11 +96,11 @@ define(["dojo/_base/declare",
...
-         if (this.docListRenderer != null && 
-             payload != null && 
-             payload.value != null &&
+         if (this.docListRenderer && 
+             payload && 
+             payload.value &&
              !isNaN(payload.value) &&
-             this.columns != payload.value)
+             this.columns !== payload.value)
          {
```
MartinDoyleUK: Lots of repetition here. Might be better with a separate variable first:

```
var newNumCols = payload && !isNaN(payload.value) && payload.value;
if (this.docListRenderer && newNumCols && newNumCols !== this.columns) {
  [log]
  [update]
}
```

Also, this will fail if the new number of columns is zero. That seems reasonable, but can you confirm?

draperd:
I've made the update as suggested... I'm ok with the 0 columns thing, because that wouldn't make any sense to request that anyway.

PR #217, opened 2015-05-06
AKU-19: Upgrade to Dojo 1.10.4

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-19 in moving Aikau up to use Dojo 1.10.4.  The unit tests revealed some issues following the upgrade which have been resolved in this PR. Unit tests are now all passing. I've also done a sniff test against Share and it also seems to be functioning as expected.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
+++ b/aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
@@ -321,7 +321,7 @@ define(["dojo/_base/declare",
             // NOTE: This line is specifically required to support Firefox, for some reason the standard
             //       key handling is being suppressed, this was uncovered on the move from Dojo 1.9.0 to
             //       both 1.9.6 and then 1.10.4
-            query(".alfresco-forms-controls-BaseFormControl .control input").on("keypress", lang.hitch(this, this.onValueEntryKeyPress));
+            query(".alfresco-forms-controls-BaseFormControl .control input", this.formWidget.domNode).on("keypress", lang.hitch(this, this.onValueEntryKeyPress));
          }
          return this.formWidget;
       },
--- a/aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
+++ b/aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
@@ -50,12 +50,13 @@ define(["dojo/_base/declare",
         "dojo/dom-attr",
         "dojo/keys",
         "dojo/_base/event",
+        "dojo/query",
         "service/constants/Default",
         "alfresco/forms/Form",
         "alfresco/forms/controls/DojoValidationTextBox",
         "alfresco/forms/controls/HiddenValue"], 
         function(declare, Property, _OnDijitClickMixin, CoreWidgetProcessing, _PublishPayloadMixin, KeyboardNavigationSuppressionMixin,
-                 template, lang, array, on, domClass, html, domAttr, keys, event) {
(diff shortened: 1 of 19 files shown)


davidcognite:
I'm really excited by this. I've had a quick review and there seem to be thankfully few changes needed. :-) I'll check this branch out and run some RM tests against it.

davidcognite:
There is a minor issue with RM - our extension of the dialog service doesn't work with the refactored code (I've not been able to trace why yet). We should be able to resolve that by switching to the Aikau dialog service, but it does mean this commit breaks backwards compatibility in some cases...

draperd:
If you're creating a dialog directly and are using the "textContent" attribute, then you'll need to change this to "content". I tried unsuccessfully to work around this issue, but you're right in that ideally you should be using the alfresco/services/DialogService to create dialogs (in which case the "textContent" attribute is honoured).

davidcognite:
We're using the DialogService, but we've extended it (prior to support for stacked dialogs) and that's the problem: https://github.com/Alfresco/records-management/blob/master/rm-share/source/web/js/alfresco/rm/services/AlfRmDialogService.js

draperd:
OK, well... we need to know whether or not this is something you can work around (especially now that the DialogService supports stacked dialogs).

aikau/src/main/resources/alfresco/services/NotificationService.js
```diff
@@ -84,7 +84,7 @@ define(["dojo/_base/declare",
             if (message) {
                this.alfPublish("ALF_CREATE_DIALOG_REQUEST", {
                   dialogId: "NOTIFICATION_PROMPT",
-                  dialogTitle: this.message("notification.prompt.title"),
+                  dialogTitle: payload.title || this.message("notification.prompt.title"),
                   textContent: message,
```
MartinDoyleUK: Do we definitely not want `dialogTitle: this.message(payload.title || "notification.prompt.title"),` here?

aikau/src/test/resources/testApp/js/aikau/testing/templates/MockXhr.html
```diff
@@ -1,4 +1,4 @@
-<div class="alfresco-testing-MockXhr alfresco-testing-MockXhr--hide-body" data-dojo-attach-point="containerNode">
+<div class="alfresco-testing-MockXhr alfresco-testing-MockXhr" data-dojo-attach-point="containerNode">
    <table class="log">
```
MartinDoyleUK: If we're going to change this, we need to remove `alfresco-testing-MockXhr--hide-body` entirely, as there are otherwise two identical classnames in there now.

aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
```diff
@@ -317,6 +318,10 @@ define(["dojo/_base/declare",
                   widgets: [primaryFormWidget].concat(autoSetFields)
                }
             }, this.formWidgetNode);
+            // NOTE: This line is specifically required to support Firefox, for some reason the standard
+            //       key handling is being suppressed, this was uncovered on the move from Dojo 1.9.0 to
+            //       both 1.9.6 and then 1.10.4
+            query(".alfresco-forms-controls-BaseFormControl .control input").on("keypress", lang.hitch(this, this.onValueEntryKeyPress));
          }
```
MartinDoyleUK: This line can potentially target controls other than just the InlineEditProperty, given the query. Is this intended, and if so, is this the best place for the code?

davidcognite:
Dave: Yes, I"m sure we can work around the issues.

PR #219, opened 2015-05-07
AKU-279: TabContainer updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-279. I've removed the JSHint errors from the code, added the missing capabilities, rewritten the existing unit test to the expected style and added a new test that loosely mirrors the use case requested by the Integrations team.

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
+++ b/aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
@@ -66,7 +66,7 @@
  *             id: "TAB1",
  *             name: "alfresco/logo/Logo",
  *             title: "Tab with Alfresco Logo",
- *             selected
+ *             selected: true,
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
```diff
@@ -18,12 +18,146 @@
...
+ * {
+ *    name: "alfresco/layout/AlfTabContainer",
+ *    config: {
+ *       widgets: [
+ *          {
+ *             id: "TAB1",
+ *             name: "alfresco/logo/Logo",
+ *             title: "Tab with Alfresco Logo",
+ *             selected
+ *             config: {
```
MartinDoyleUK: this doesn't look like valid JSON?

PR #221, opened 2015-05-08
AKU-244: Sanitize HTML proxy WebScript

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-244. 

I've created a new proxy WebScript that accepts URIs for Alfresco Repository REST APIs and runs a single property of each item through the stripUnsafeHTML() function. This is currently limited in that it only sanitizes a single attribute attribute and nested attributes can't be used with dot-notation properties.

However, I think we can live with these limitations initially and iterate as the use-cases arise. I've done manual testing to verify it works, but have not provided an automated test because we have no satisfactory way of testing WebScripts at the moment.

The CommentService has been updated to make use of this new proxy WebScript and I have verified that it works as expected.

### DIFF
--- a/aikau/src/main/resources/webscripts/sanitize-response.get.js
+++ b/aikau/src/main/resources/webscripts/sanitize-response.get.js
@@ -1,7 +1,4 @@
-/* global args */
-/* global remote */
-/* global stringUtils */
-/* global jsonUtils */
+/*global args,remote,stringUtils,jsonUtils*/
 var response = {};
 if (args.ws)
 {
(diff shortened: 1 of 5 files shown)


aikau/src/main/resources/webscripts/sanitize-response.get.js
```diff
@@ -0,0 +1,35 @@
+/* global args */
+/* global remote */
+/* global stringUtils */
+/* global jsonUtils */
+var response = {};
```
MartinDoyleUK: Can write all of above as `/*global args,remote,stringUtils,jsonUtils*/`

aikau/src/main/resources/webscripts/sanitize-response.get.js
```diff
@@ -0,0 +1,35 @@
...
+if (args.ws)
+{
+   var connector = remote.connect("alfresco");
+   var json = connector.get(encodeURI(args.ws));
+   try
+   {
+      // PLEASE NOTE: Very deliberate usage of == in the following line
```
MartinDoyleUK: Shouldn't be necessary with JSHint comment below

PR #222, opened 2015-05-11
AKU-278: Add "striped content" widget

MartinDoyleUK:
This addresses issue [AKU-278](https://issues.alfresco.com/jira/browse/AKU-278) which is a feature request for a new style of layout widget called "striped content", which is a stylable, full-width container with a centred, fixed-width content.

### DIFF
--- a/aikau/src/main/resources/alfresco/css/less/defaults.less
+++ b/aikau/src/main/resources/alfresco/css/less/defaults.less
@@ -1,3 +1,17 @@
+// Main palette
+@primaryBG: #fff;
+@primaryFG: #333;
+@headerBG: #0082c8;
+@headerFG: #fff;
+@alternateBG: #ccc;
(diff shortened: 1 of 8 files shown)


aikau/src/main/resources/alfresco/css/less/defaults.less
```diff
@@ -1,3 +1,17 @@
+// Main palette
+@primaryBG: #fff;
+@primaryFG: #333;
+@headerBG: #0082c8;
```
draperd: This fine for now, however - with AKU-286 we might need to change these to be black to match the current header colour.

MartinDoyleUK: Theme-wise, it'd be good to have it as off-black, because it's a bit more stylish. I don't know why. Just off-black is the new black :)

PR #224, opened 2015-05-11
AKU-203:Document locate action

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-203. This isn't quite complete pending work on AKU-162 (for link creation) so this service will be updated when link creation is supported.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/_ItemLinkMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_ItemLinkMixin.js
@@ -63,18 +63,22 @@ define(["dojo/_base/declare",
        *
        * @instance
        * @param {object} publishPayload The publication payload object to update.
+       * @param {object} item An optional item to provide. If this is not provided then "currentItem" is used instead
        */
-      updateFolderLinkPublication: function alfresco_renderers__ItemLinkMIxin__updateFolderLinkPublication(publishPayload) {
-         var locn = this.currentItem.location;
+      updateFolderLinkPublication: function alfresco_renderers__ItemLinkMIxin__updateFolderLinkPublication(publishPayload, item) {
+         item = item || this.currentItem;
+         var locn = item.location;
          publishPayload.path = this.combinePaths(locn.path, locn.file);
       },
 
       /**
        * @instance
        * @param {object} payload The payload to add the link data to
+       * @param {object} item An optional item to provide. If this is not provided then "currentItem" is used instead
        * @returns {string} The topic to publish on
        */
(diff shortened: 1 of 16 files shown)


aikau/src/main/resources/alfresco/renderers/_ItemLinkMixin.js
```diff
@@ -82,10 +86,10 @@ define(["dojo/_base/declare",
          }
          else
          {
-            if (this.currentItem && this.currentItem.node)
+            if (item && item.node)
             {
-               var jsNode = this.currentItem.jsNode;
-               if (jsNode.isLink && this.currentItem.location.site)
+               var jsNode = item.jsNode;
+               if (jsNode.isLink && item.location.site)
```
MartinDoyleUK: Could item.jsNode be `null` or `undefined` here (i.e. causing `jsNode.isLink` to error)?

aikau/src/main/resources/alfresco/services/actions/NodeLocationService.js
```diff
@@ -0,0 +1,163 @@
...
+       * @default false
+       */
+      useAikauPages: false,
+
+      /**
+       * @instance
+       * @param {array} args Constructor arguments
+       */
+      constructor: function alfresco_services_actions_NodeLocationService__constructor(args) {
+         lang.mixin(this, args);
```
MartinDoyleUK: Should we be using [`dojo/_base/declare#safeMixin()`](https://dojotoolkit.org/reference-guide/1.10/dojo/_base/declare.html#dojo-base-declare-safemixin) here instead?

aikau/src/test/resources/alfresco/services/actions/NodeLocationTest.js
```diff
@@ -0,0 +1,114 @@
...
+            type: "PUBLISH",
+            topic: "ALF_NAVIGATE_TO_PAGE",
+            pos: "last"
+         })
+         .then(function(payload) {
+            assert.isNotNull(payload, "Navigation publication not found");
+            if (payload) {
+               assert.equal(payload.url, "/aikau/page/site/site1/documentlibrary#path=%2Fsome%2Frandom%2Fpath", "Unexpected location");
+            }
+         });
```
MartinDoyleUK: /like

aikau/src/main/resources/alfresco/services/actions/NodeLocationService.js
draperd: Not sure, I've used lang.mixin in all the other service constructors as well. We'd probably want to change all of them.

PR #229, opened 2015-05-13
AKU-282: ActionService updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-282


aikau/src/main/resources/alfresco/navigation/_HtmlAnchorMixin.js
```diff
@@ -42,7 +42,7 @@ define(["dojo/_base/declare",
         "dojo/query",
         "dojo/NodeList",
         "dojo/NodeList-manipulate"], 
-        function(declare, AlfCore, _NavigationServiceTopicMixin, AlfConstants, lang, array, query, NodeList, NodeListManipulate) {
+        function(declare, AlfCore, _NavigationServiceTopicMixin, AlfConstants, lang, array, query /*, NodeList, NodeListManipulate*/) {
```
MartinDoyleUK: Why not delete? That would be the norm, I believe?

draperd: The reason is that the dependencies aren't required, but they aren't referenced. I could have JSHinted them out, but I wanted to make sure that if someone adds in additional dependencies that they don't end up with a mis-match between the array of dependencies and the callback args

PR #232, opened 2015-05-13
Add some error handling to Service loading method

davidcognite:
This code throws a better error message if the service wasn't loaded properly. This can be considered part of the work to prevent silent failures.

Previous error was: "BOOM! number is not a function"
New error is: "BOOM! Unable to locate service: aikauTesting/mockservices/404"

I think there's more we could do to let the user know of require failures. I'll pop a JIRA feature request in the queue.

Note: No tests added, as I'm not sure what tests you can do for a page that throws a JS error on load.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/Page.js
+++ b/aikau/src/main/resources/alfresco/core/Page.js
@@ -20,7 +20,7 @@
 /**
  * This is used as the default root object when instantiating a page. There should be no need
  * to ever instantiate this widget directly.
- * 
+ *
  * @module alfresco/core/Page
  * @extends module:alfresco/core/ProcessWidgets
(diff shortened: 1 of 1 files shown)


draperd:
Before anyone merges this, I was actually planning on looking at a lot of these issues under AKU-284 (which was why I increased the size to give some scope to general error handling improvements). I appreciate this is a minor change, but I want to rework the general error handling anyway. We can discuss this further if you like though.

davidcognite:
Ok, that's fair enough. I'll close this PR and open it up again on a different local branch (I've got another PR waiting on my local develop branch). I agree that error handling for service/widget loading should be part of a wider sweep through the processing code to prevent hiding errors.

PR #246, opened 2015-05-21
Prototype HtmlListView

draperd:
PLEASE NOTE: Martin and David - do not pull this request (I'm just using this PR for education purposes).

This PR contains an example of how you can extend the current "alfresco/lists/views/AlfListView" and "alfresco/lists/views/ListRenderer" widgets to create a view that does present the list in a tabular form or use HTML table related elements. 

Instead it overrides base widgets and adds in new templates for rendering items. I've included a unit test page that demonstrates how this could be used. If you checkout the branch related to this PR and start the unit test application you can try it out at:  http://localhost:8089/aikau/page/tp/ws/HtmlListView

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/views/HtmlListView.js
+++ b/aikau/src/main/resources/alfresco/lists/views/HtmlListView.js
@@ -1,0 +1,99 @@
+/**
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * 
+ * @module alfresco/lists/views/HtmlListView
+ * @extends module:alfresco/lists/views/AlfListView
+ * @author Dave Draper
+ */
+define(["dojo/_base/declare",
+        "alfresco/lists/views/AlfListView",
+        "dojo/text!./templates/HtmlListView.html",
+        "alfresco/lists/views/ListRenderer",
+        "dojo/text!./templates/HtmlListViewRenderer.html",
+        "dojo/_base/lang",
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/lists/views/HtmlListView.js
```diff
@@ -0,0 +1,99 @@
...
+        "alfresco/lists/views/AlfListView",
+        "dojo/text!./templates/HtmlListView.html",
+        "alfresco/lists/views/ListRenderer",
+        "dojo/text!./templates/HtmlListViewRenderer.html",
+        "dojo/_base/lang",
+        "dojo/_base/array",
+        "dojo/dom-construct"], 
+        function(declare, AlfListView, template, ListRenderer, rendererTemplate, lang, array, domConstruct) {
+   
+   var Renderer = declare([ListRenderer], {
```
draperd: Note: I've created another module within the HtmlListView - it would be better for re-use if I'd abstracted this to a separate module, but for brevity I've just added it here. When we finish this off, it would be in its own module.

aikau/src/main/resources/alfresco/lists/views/HtmlListView.js
```diff
@@ -0,0 +1,99 @@
...
+
+      templateString: rendererTemplate,
+
+      renderData: function alfresco_dnd_DragAndDropItems__renderData() {
```
draperd: I've got the function name wrong here, because I used an existing renderer as a sample, whoops :)

aikau/src/main/resources/alfresco/lists/views/HtmlListView.js
```diff
@@ -0,0 +1,99 @@
...
+         array.forEach(this.items, function(item) {
+            var property = lang.getObject(this.propertyToRender, false, item);
```
draperd: We iterate over the items here that are set in the main HtmlListView widget - all of the data handling, etc. is handled by the parent module, we're just overriding what to actually do with each item. I've used a closure here - but once again - when this is finalized in a future sprint it would be a bound function.

aikau/src/main/resources/alfresco/lists/views/HtmlListView.js
```diff
@@ -0,0 +1,99 @@
...
+       * Creates a new [ListRenderer]{@link module:alfresco/lists/views/ListRenderer}
+       * which is used to render the actual items in the view. This function can be overridden by extending views
+       * (such as the [Film Strip View]{@link module:alfresco/documentlibrary/views/AlfFilmStripView}) to create
+       * alternative widgets applicable to that view.
+       * 
+       * @instance
+       * @returns {object} A new [ListRenderer]{@link module:alfresco/lists/views/ListRenderer}
+       */
+      createListRenderer: function alfresco_dnd_DragAndDropItemsListView__createListRenderer() {
+         var dlr = new Renderer({
```
draperd: This is where we create the Renderer declared further above, we could pass in any data we wanted here, but I'm just passing the data to render along with a configurable property to select from each item.

PR #250, opened 2015-05-22
AKU-293, AKU-302: Pagination using hash issues.

draperd:
This PR addresses issues raised by both https://issues.alfresco.com/jira/browse/AKU-293 and https://issues.alfresco.com/jira/browse/AKU-302 which reported issues that occur when trying to set useHash to true and then loading pages with currentPage and currentPageSize harh parameters.

One of the issues was related to the _AlfHashMixin which was recently update to provide more advanced support for managing hash variables and using those values to update the hash. However, we need to be able to sort the "core" hash behaviour of the list widgets (including pagination) such that additional hash relative behaviour can be specified if requested.

I've added a new unit test to ensure that loading a page containing a paginated list with hashing enabled works correctly. I've also made various other minor tweaks that I encountered whilst working with other modules.


aikau/src/main/resources/alfresco/documentlibrary/_AlfHashMixin.js
```diff
@@ -148,25 +148,41 @@ define(["dojo/_base/declare",
...
+            // Iterate over the keys defined in hashVarsForUpdate - return true if the payload contains one of them
+            for(var i=0; i < this.hashVarsForUpdate.length; i++)
             {
-               if (updateInstanceValues === true)
+               if(this.hashVarsForUpdate[i] in payload)
                {
-                  this[this.hashVarsForUpdate[i]] = payload[this.hashVarsForUpdate[i]];
+                  if (updateInstanceValues === true)
+                  {
+                     this[this.hashVarsForUpdate[i]] = payload[this.hashVarsForUpdate[i]];
```
MartinDoyleUK: Do we want to delete values if we get undefined/null coming through, or are we happy just persisting exactly what we receive?

aikau/src/main/resources/alfresco/documentlibrary/_AlfHashMixin.js
```diff
@@ -148,25 +148,41 @@ define(["dojo/_base/declare",
...
+            // AKU-293 and AKU-302. We want to be able to support updates for hash parameters that
+            // do NOT require updating instance values. At some point we should be able to remove
+            // this code block with refactoring to the AlfSearchList and AlfDocumentList
+            if (this._coreHashVars)
+            {
+               for(i=0; i < this._coreHashVars.length; i++)
+               {
+                  if(this._coreHashVars[i] in payload)
+                  {
+                     containsUpdateableVar = true;
```
MartinDoyleUK: Even if the property is null/undefined?

PR #269, opened 2015-06-01
AKU-316: Update forms to support hashVarsForUpdate

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-316 in updating alfresco/forms/Form to support the hashVarsForUpdate property defined in the alfresco/documentlibrary/_AlfHashMixin. This means that it's now possible to configure a form so that only some form control values are updated from the hash and that only certain form control values update the hash.


davidcognite:
Looks good to me.

PR #270, opened 2015-06-01
AKU-317: adds prevent cache support to serviceXhr

davidcognite:
Also adds support to CrudService for it.


davidcognite:
I wasn't sure what to do about tests for this as coreXhr lacks specific tests.

draperd:
I'm assuming you ran the existing tests though (as many of those go through CoreXhr functions)?

davidcognite:
I realised that I could extend the CrudService tests to cover both my initial commits, so all tested now.

PR #279, opened 2015-06-04
Adding library version information to the manifest.mf

covolution:
This change adds library information (available at build time) to the Jar manfest file. It just lists surf, dojo, jquery, jqueryui by using dependency properties in the pom.

This should make it easier to upgrade those libraries by just changing the property. However, I notice that jquery/jquery ui are saved in the source tree whereas dojo is a dependency.  That would require some rework.

### DIFF
--- a/aikau/pom.xml
+++ b/aikau/pom.xml
@@ -134,7 +134,7 @@
                         <artifactItem>
                            <groupId>org.dojotoolkit</groupId>
                            <artifactId>dojo</artifactId>
-                           <version>1.10.4</version>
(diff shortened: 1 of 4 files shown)


draperd:
This doesn't actually address all the files where those properties are used...  aikau-extension.xml and the testApp/WEB-INF/surf.xml also reference Dojo, JQuery etc... any chance you can update the PR to deal with those files as well?

covolution:
I had already included  aikau-extension.xml but not the surf.xml from the testApp.  I added this now.

PR #281, opened 2015-06-04
AKU-331: Updated CrudService to encode URIs by default

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-331 to update the alfresco/services/CrudService to encode all URIs before they're requested (however, it's possible to configure the service so that URIs aren't encoded). A unit test has been added to verify encoding occurs.


davidcognite:
The CrudService feels like the wrong place for this fix. Would support for URI encoding be better in the serviceXhr method, rather than in each service?

draperd:
I've pushed an update, can you re-review please?

PR #285, opened 2015-06-05
Current Document Library evaluation and updates to ensure it all works

draperd:
Just spent a bit of time working through the current Document Library to make sure we hadn't regressed anything and took the opportunity to tidy a few things up.

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/AlfBreadcrumb.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/AlfBreadcrumb.js
@@ -102,7 +102,7 @@ define(["dojo/_base/declare",
       postCreate: function alfresco_documentlibrary_AlfBreadcrumbTrail__postCreate() {
          if (this.label)
          {
-            this.breadcrumbNode.innerHTML = this.encodeHTML(this.message(this.label));
+            this.breadcrumbNode.innerHTML = this.message(this.label);
          }
       },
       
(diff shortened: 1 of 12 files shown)


draperd:
Don't pull this yet - I need to push a couple of updates to unit tests that are failing.

draperd:
Pushed the required test updates

aikau/src/main/resources/alfresco/documentlibrary/AlfBreadcrumb.js
```diff
@@ -102,7 +102,7 @@ define(["dojo/_base/declare",
       postCreate: function alfresco_documentlibrary_AlfBreadcrumbTrail__postCreate() {
```
davidcognite: Name shouldn't contain "Trail"

aikau/src/main/resources/alfresco/documentlibrary/AlfBreadcrumb.js
```diff
@@ -102,7 +102,7 @@ define(["dojo/_base/declare",
...
          if (this.label)
          {
-            this.breadcrumbNode.innerHTML = this.encodeHTML(this.message(this.label));
+            this.breadcrumbNode.innerHTML = this.message(this.label);
```
davidcognite: Is the label sanitised elsewhere?

draperd: The filter publications are currently only issues by filters that have to be manually added to page, so there is minimal rish. The issue here was that when clicking on any of the tag filters we were getting double encoding of quotes (which is why I've removed the encoding).

PR #287, opened 2015-06-08
AKU-344: Dialog containing PublishingDropDownMenu close

draperd:
This PR addresses comments in https://issues.alfresco.com/jira/browse/AKU-344 relating to alfresco/renderers/PublishingDropDownMenu widgets not detecting cancellation actions when a dialog is closed (by clicking the close icon rather than using the ESCAPE key).

This change "normalises" behaviour by simulating an ESCAPE key press when the dialog closes to ensure that anything listening to the escape key is notified. In order to efficiently write the test for this I've made some updates to the way that iterated data (e.g. in lists) is handled so that IDs can be applied more effectively. This will go some way to addressing https://issues.alfresco.com/jira/browse/AKU-353.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/views/layouts/_LayoutMixin.js
+++ b/aikau/src/main/resources/alfresco/lists/views/layouts/_LayoutMixin.js
@@ -1,0 +1,48 @@
+/**
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @module alfresco/lists/views/layouts/_LayoutMixin
+ * @extends module:alfresco/core/CoreWidgetProcessing
+ * @author Dave Draper
+ */
+define(["dojo/_base/declare",
+        "alfresco/core/CoreWidgetProcessing"], 
+        function(declare, CoreWidgetProcessing) {
+   
+   return declare([CoreWidgetProcessing], {
+
+      /**
+       * Extends the [inherited function]{@link module:alfresco/core/CoreWidgetProcessing#processWidgetConfig}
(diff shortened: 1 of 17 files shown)


aikau/src/main/resources/alfresco/lists/views/layouts/_MultiItemRendererMixin.js
```diff
@@ -41,10 +41,11 @@ define(["dojo/_base/declare",
         "dojo/_base/array",
         "dojo/_base/lang",
         "dojo/dom-style",
+        "dojo/dom-attr",
         "dojo/on",
         "dojo/_base/event"], 
         function(declare, CoreWidgetProcessing, ObjectProcessingMixin, ObjectTypeUtils, JsNode, _AlfDocumentListTopicMixin, 
-                 domClass, array, lang, domStyle, on, event) {
+                 domClass, array, lang, domStyle, domAttr, on, event) {
```
davidcognite: You've added domAttr, but I can't see where it's actually being used.

aikau/src/main/resources/alfresco/lists/views/layouts/_LayoutMixin.js
```diff
@@ -0,0 +1,48 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @module alfresco/lists/views/layouts/_LayoutMixin
```
davidcognite: Should there be a module level JSDoc description?

aikau/src/main/resources/alfresco/lists/views/layouts/_LayoutMixin.js
```diff
@@ -0,0 +1,48 @@
...
+       *
+       * @instance
+       * @param {object} widget The widget configuration build configuration for
+       * @return {object} The arguments that can be used when instantiating the widget configuration processed
+       */
+      processWidgetConfig: function alfresco_lists_views_layouts_Cell__processWidgetConfig(widget) {
+         if (widget.id)
+         {
+            widget.id = widget.id + "_ITEM_" + this.currentItem.index;
+         }
```
davidcognite: if there's no id (or maybe anyway?), would it be useful to add a class of e.g. ITEM_0 to be able to select all widgets for the item?

draperd: I see what you're saying here... but in the common scenario where there is more than one type of renderer or layout control for a single item you're almost immediately going to hit duplicates. If you prefix with a generated ID then it won't be useful for testing either. If this forces people to actually include IDs on their models - then so much the better!

davidcognite: The duplicates was why I was suggesting a class, it'd give us a different way to access the widgets.  But the point about the ids is a good one...

PR #288, opened 2015-06-09
AKU-336: Annotating JSDoc with support statements

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-336 and defines some new JSDoc tags that can be used to indicate which modules and functions will be supported in Aikau going forwards. An initial set of modules have been updated to use these tags. This is not necessarily the final implementation but we'll need to review at the sprint end to see if we're happy with the approach going forwards.

### DIFF
--- a/aikau/src/jsdoc-plugins/customization-support-tags.js
+++ b/aikau/src/jsdoc-plugins/customization-support-tags.js
@@ -1,0 +1,69 @@
+/* global exports */
+exports.defineTags = function(dictionary) {
+
+   // When something is page safe it can be used in a page model in the knowledge that it isn't
+   // going to be removed or significantly changed in a future release (this should be applied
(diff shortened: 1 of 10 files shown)


aikau/src/jsdoc-plugins/customization-support-tags.js
```diff
@@ -0,0 +1,69 @@
+/* global exports */
+exports.defineTags = function(dictionary) {
+
+   // When something is page safe it can be used in a page model in the knowledge that it isn't
+   // going to be removed or significantly changed in a future release (this should be applied
+   // to widgets and services)
+   dictionary.defineTag("pageSafe", {
```
davidcognite: These tag definitions should also be on a wiki page, or somewhere similar.

PR #292, opened 2015-06-10
AKU-354: Sites menu updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-354 to ensure that the asynchronously loaded favourites cascading menu is hidden before being populated and then redisplayed to ensure that the popup positioning and scrolling are consistent each time it is used (the asynchronous loading only takes place the first time the cascade is opened). I've not updated the unit test for this as the existing tests prove that the menu can be opened and the contents displayed.

### DIFF
--- a/aikau/src/main/resources/alfresco/header/AlfSitesMenu.js
+++ b/aikau/src/main/resources/alfresco/header/AlfSitesMenu.js
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2005-2013 Alfresco Software Limited.
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
  *
  * This file is part of Alfresco
  *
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/header/AlfSitesMenu.js
```diff
@@ -382,18 +382,21 @@ define(["dojo/_base/declare",
...
+         // Close the popup before adding favourites so that it can be re-opened once populated
+         // to ensure consistency of popup placement (see AKU-354)...
+         popup.close(this.favoritesCascade.popup);
+
          // Check for keyboard access by seeing if the first child is focused...
          var focusFirstChild = (this.favouritesList && this.favouritesList.getChildren().length > 0 && this.favouritesList.getChildren()[0].focused);
          
          // Remove the loading favourites item...
-         var _this = this;
          array.forEach(this.favouritesList.getChildren(), function(widget, index) {
```
MartinDoyleUK: `index` not necessary here

PR #293, opened 2015-06-12
AKU-299: Updates to support PubSub DND clear

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-299 to make it possible to clear alfresco/forms/controls/DragAndDropTargetControl widgets via PubSub. There are two variations of clear action - one which does a brute force clear, the second which clears items in such a way that single-use items are returned to their palette. I've also refactored out the DND unit tests to make the easier to run individually


aikau/src/main/resources/alfresco/forms/controls/DragAndDropTargetControl.js
```diff
@@ -120,6 +147,58 @@ define(["dojo/_base/declare",
...
+      onClearDroppedItems: function cmm_forms_controls_DragAndDropTargetControl__onClearDroppedItems() {
+         var nodes = lang.getObject("wrappedWidget.previewTarget.map", false, this);
+         for (var node in nodes)
+         {
+            if (nodes.hasOwnProperty(node))
+            {
+               var widget = registry.byId(node);
+               if(widget && typeof widget.onItemDelete === "function")
+               {
+                  widget.onItemDelete.apply(widget);
```
MartinDoyleUK: `widget.onItemDelete();` would be simpler

PR #294, opened 2015-06-12
AKU-332: Identify empty results with CSS

MartinDoyleUK:
This PR addresses issue [AKU-332](https://issues.alfresco.com/jira/browse/AKU-332), which requests CSS class identification of certain list elements/messages, in order to make them more testable/stylable.


aikau/src/main/resources/alfresco/lists/views/templates/AlfListView.html
```diff
@@ -1,3 +1,3 @@
-<div class="alfresco-lists-views-AlfListView">
+<div class="${baseClass}">
```
draperd: If someone changes baseClass through config, isn't that going to totally break the widget? Wouldn't it be better to augment the "alfresco-lists-view-AlfListView" class with additional classes? In which case they could just use "additionalCssClasses" as that is now supported for all widgets in a standard model.

PR #298, opened 2015-06-15
AKU-356: Auto-publish form on field changes

MartinDoyleUK:
This addresses issue [AKU-356](https://issues.alfresco.com/jira/browse/AKU-356), and permits the auto-publishing of a form when its values change. By default, it will not publish when any of its values are invalid, but this can be turned on with a config option. For full usage information, see the JSDoc.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/Form.js
+++ b/aikau/src/main/resources/alfresco/forms/Form.js
@@ -290,7 +290,7 @@ define(["dojo/_base/declare",
             valid: isValid,
             invalidFormControls: this.invalidFormControls
          });
-         if(this.autoSavePublishTopic !== null && (isValid || this.autoSaveOnInvalid)) {
+         if(this.autoSavePublishTopic && typeof this.autoSavePublishTopic === "string" && (isValid || this.autoSaveOnInvalid)) {
             this.alfPublish(this.autoSavePublishTopic, this.autoSavePublishPayload || this.getValue(), this.autoSavePublishGlobal);
          }
       },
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -515,7 +585,7 @@ define(["dojo/_base/declare",
        * @instance
        */
       createButtons: function alfresco_forms_Form__createButtons() {
-         if (this.showOkButton === true)
+         if (this.showOkButton === true && this.autoSavePublishTopic === null)
```
draperd: Confused by this... why not !this.autoSavePublishTopic... having the empty string as a valid topic wouldn't make a lot of sense.

aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -548,7 +618,7 @@ define(["dojo/_base/declare",
             domConstruct.destroy(this.okButtonNode);
          }
          
-         if (this.showCancelButton === true)
+         if (this.showCancelButton === true && this.autoSavePublishTopic === null)
```
draperd: See previous comment

aikau/src/test/resources/alfresco/forms/controls/BaseFormTest.js
```diff
@@ -41,75 +42,116 @@ define(["intern!object",
          browser.end();
       },
 
-      // teardown: function() {
```
draperd: We can remove these commented out blocks now when we see them

MartinDoyleUK: It has been removed. That's why it's in red ...

draperd: Oh, yes... silly me :)

PR #301, opened 2015-06-15
Form CSS improvements. 

kevinroast:
Left-to-right field row rendering. Minor font size improvements for Alfresco Share consistency.

Layout changes only - no functional modifications.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/css/BaseFormControl.css
+++ b/aikau/src/main/resources/alfresco/forms/controls/css/BaseFormControl.css
@@ -16,8 +16,12 @@
 .alfresco-forms-controls-BaseFormControl > .description-row > .description {
    font-size: @small-font-size;
    line-height: ceil(@small-font-size/@standard-line-height) * @standard-line-height;
-   margin: 0 0 @standard-line-height @dialog-label-section + 4;
+   margin: 0 0 @standard-line-height 0;
    color: @de-emphasized-font-color;
+}
+
+.alfresco-dialog-AlfDialog .dialog-body .alfresco-forms-controls-BaseFormControl > .description-row > .description {
+   margin: 0 0 @standard-line-height @dialog-label-section + 4;
 }
 
 .alfresco-forms-controls-BaseFormControl > .control-row {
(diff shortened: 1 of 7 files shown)


draperd:
I've checked out the latest changes on your branch and ran the full test suite, I'm getting the following error:

Text Box Tests
- Check validation error message
  "Validation error message should be displayed"

(FYI... You can target specific test suites by updating the Suites.js file - there's a section at the top we use for individual tests, just switch which version of "baseFunctionalSuites" has the "x" in front of it and set the the unit test as the entry in the array).

draperd:
Also, there's still an issue with the validation icon being to the left on page forms (when it used to be on the right). I'm not sure if we want to worry about that or not.

kevinroast:
>I've checked out the latest changes on your branch and ran the full test suite, I'm getting the following error:
>Text Box Tests
>Check validation error message "Validation error message should be displayed"

Because of course the test checks for a specific CSS class, which is no longer present because now the Form CSS is only applied in dialog. I've switched it back. Perhaps the test can be made less brittle? I've had to change it twice now and each time it has not pointed out an error at all.

kevinroast:
>Also, there's still an issue with the validation icon being to the left on page forms (when it used to be on the right). I'm not sure if we want to worry about that or not.
That's not an issue.

aikau/src/main/resources/alfresco/forms/controls/css/BaseFormControl.css
```diff
@@ -1,37 +1,31 @@
...
 }
 
-.alfresco-forms-controls-BaseFormControl > .title-row > label {
-   font-size: @normal-font-size;
-   line-height: ceil(@normal-font-size/@standard-line-height) * @standard-line-height;
-   margin-bottom: 3px;
-   display: inline-block;
-}
-
-.alfresco-forms-controls-BaseFormControl.hiddenlabel > .title-row {
```
draperd: You've removed the .hiddenLabel class but you've not removed all the references to it, e.g. from PdfJs.js, InlineEditProperty.js, InlineEditSelect.js, Tags.js. If we're removing the class we should probably also be removing references to it.

kevinroast:
I haven't removed it... it's just moved down the file..!
+.alfresco-forms-controls-BaseFormControl.hiddenlabel > .title-row {
+   position: absolute;
+   left: -999em;
+   height: 0;
+}

PR #305, opened 2015-06-16
Update Vagrant

MartinDoyleUK:
Simplifies and updates the vagrant setup. Uses the latest versions of Chrome, Firefox, Selenium and Ubuntu. Has appropriate machine spec defaults for use with VirtualBox VMs.


kevinroast:
Working good for me now with Chrome on Linux with that change - FYI needed to update vagrant to latest 1.7.2 (was on 1.4.3)

PR #317, opened 2015-06-17
AKU-66: SearchListTest updates

draperd:
This fixes up the test failures caused by the previous PR for https://issues.alfresco.com/jira/browse/AKU-66

### DIFF
--- a/aikau/src/test/resources/alfresco/documentlibrary/SearchListTest.js
+++ b/aikau/src/test/resources/alfresco/documentlibrary/SearchListTest.js
@@ -77,8 +77,8 @@ define(["intern!object",
          // Check that updating the hash results in a search request being made...
          .getLastPublish("ALF_SEARCH_REQUEST")
          .then(function(payload) {
-            assert.equal(payload.term, "testTerm1", "Search term didn't request search");
-            assert.equal(payload.facetFields, "qname1", "Facet fields not set appropriately from request");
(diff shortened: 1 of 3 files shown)


aikau/src/test/resources/alfresco/documentlibrary/SearchListTest.js
```diff
@@ -75,79 +75,68 @@ define(["intern!object",
...
-      "Test that facet fields are included in search": function() {
-         // Make sure our facet was included...
-         return browser.findAllByCssSelector(TestCommon.pubSubDataCssSelector("last", "facetFields", "qname1"))
-            .then(function(elements) {
-               assert(elements.length === 1, "Facet fields not set appropriately from request");
-            });
+         .getLastPublish("ALF_SEARCH_REQUEST")
+         .then(function(payload) {
+            assert.equal(payload.term, "testTerm1", "Search term didn't request search");
+            assert.equal(payload.facetFields, "qname1", "Facet fields not set appropriately from request");
```
MartinDoyleUK: Can also do `assert.propertyVal(payload, "facetFields", "qname1", "Facet fields...");`
Deep testing also available with `assert.deepPropertyVal(payload, "foo.bar.wibble", "expectedValue", "Value not correct");`
There are also corresponding negative methods on the `assert` object of `propertyNotVal()` and `deepPropertyNotVal()` with similar syntax.
But this works too, of course :)

PR #318, opened 2015-06-18
AKU-2: Added unitTestsGrid maven profile

jphuynh:
- Added unitTestsGrid maven profile
- Runtest_grid task in the Maven integration-test phase
- Unplugged startUnitTestApp task for grid as Tomcat is started during
the pre-integration-test phase by Maven.

I added a new Maven profile to avoid collision with ongoing
developments (ie: The run profile). For the same reasons, I kept all
the jetty part and only
unplugged that for the test_grid task.

Run command: mvn verify -PunitTestsGrid

@ohej , @MartinDoyleUK, could you guys review my changes and tell me if it's not colliding with your current developments around Tomcat ?

Thanks.

-Jp

### DIFF
--- a/aikau/pom.xml
+++ b/aikau/pom.xml
@@ -55,6 +55,80 @@
                </dependency>
             </dependencies>
         </profile>
+        <profile>
+            <id>unitTestsGrid</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.tomcat.maven</groupId>
+                        <artifactId>tomcat7-maven-plugin</artifactId>
+                        <version>2.2</version>
+                        <executions>
+                            <execution>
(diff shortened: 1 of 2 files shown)


aikau/pom.xml
```diff
@@ -55,6 +55,80 @@
...
+                                <phase>post-integration-test</phase>
+                                <goals>
+                                    <!-- Safely stop tomcat when integration-test phase is done -->
+                                    <goal>shutdown</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                      <artifactId>maven-antrun-plugin</artifactId>
```
jphuynh: grunt call included in the profile so it doesn't affect the integration-test phase outside the unit tests on grid context.

PR #323, opened 2015-06-18
AKU-360: SearchBox Customization Improvements

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-360 and tidies up the display when advanced search menus are hidden. It also allows custom publications to be configured for all search result clicks.


kevinroast:
Can i ask the width of the listsearch results box is no longer fixed? i.e. use a min-width not fixed width - as this was changed a while back and impairs the usability IMHO.

draperd:
I've made an additional change to ensure the width grows (up to max width) as it did previously.

aikau/src/main/resources/alfresco/header/SearchBox.js
```diff
@@ -592,6 +687,14 @@ define(["dojo/_base/declare",
             this._searchMenu.placeAt(this._searchMenuNode);
             this._searchMenu.startup();
          }
+         else
+         {
+            // When not including the advanced search menu, we'll just include an element that indicates
+            // that indicates the control is still related to search. It's purely for decoration though.
+            domConstruct.create("span", {
+               "class": "alfresco-header-SearchBox-menu--hideAdvancedSearch"
```
MartinDoyleUK: Really, this is a case of a CSS class describing what it does, not what it is. It should be something like `--advancedSearchDisabled` and then leave the CSS to decide whether to hide it, grey it out, make it smaller, etc.

PR #334, opened 2015-06-22
Tweaks in preparation for 502RC

draperd:
This PR provides some general tweaks for the next release of Aikau running in Share. Primarily a lot of this focuses on the Aikau PDF.js preview plugin working reliably.


aikau/src/main/resources/alfresco/core/CoreWidgetProcessing.js
```diff
@@ -201,7 +201,7 @@ define(["dojo/_base/declare",
          // If the widget has dynamic visibility behaviour configured then we need to set up the necessary
          // subscriptions to handle the rules that have been defined. We will set the initial visibility
          // as requested and then set up the subcriptions...
-         if (widget[configAttribute] !== undefined)
+         if (widget[configAttribute])
```
MartinDoyleUK: Could the config attribute values be falsy?

draperd: No, it will be an object or nothing.

PR #343, opened 2015-06-26
AKU-375: Twister updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-375 to update the alfresco/layout/Twister widget to support the ability to configure initial open/closed state. It also adds the ability to use the alfresco/services/PreferenceService to retrieve and store the preferred open/closed state of the twister from the current users personal preferences. For good measure I've also updated the JSDoc and fixed up the unit tests.

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/Twister.js
+++ b/aikau/src/main/resources/alfresco/layout/Twister.js
@@ -115,18 +115,18 @@ define(["dojo/_base/declare",
        *
        * @instance
        * @type {string}
-       * @default "alfresco-layout-Twister-open"
+       * @default "alfresco-layout-Twister--open"
        */
-      CLASS_OPEN: "alfresco-layout-Twister-open",
+      CLASS_OPEN: "alfresco-layout-Twister--open",
       
       /**
        * The CSS class applied when the twister is closed (and removed when opened).
        *
        * @instance
        * @type {string}
-       * @default "alfresco-layout-Twister-closed"
+       * @default "alfresco-layout-Twister--closed"
        */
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/layout/Twister.js
```diff
@@ -65,6 +111,24 @@ define(["dojo/_base/declare",
       templateString: template,
 
       /**
+       * The CSS class applied when the twister is opened (and removed when closed).
+       *
+       * @instance
+       * @type {string}
+       * @default "alfresco-layout-Twister-open"
+       */
+      CLASS_OPEN: "alfresco-layout-Twister-open",
```
MartinDoyleUK: Can this use BEM and be `alfresco-layout-Twister--open` instead?

aikau/src/main/resources/alfresco/layout/Twister.js
```diff
@@ -65,6 +111,24 @@ define(["dojo/_base/declare",
...
+      
+      /**
+       * The CSS class applied when the twister is closed (and removed when opened).
+       *
+       * @instance
+       * @type {string}
+       * @default "alfresco-layout-Twister-closed"
+       */
+      CLASS_CLOSED: "alfresco-layout-Twister-closed",
```
MartinDoyleUK: Why do we need a closed class when it can only have two states? Surely the absence of open is closed (or vice versa, if preferred)?

draperd: I'd prefer to have the specific states explicitly declared

PR #344, opened 2015-06-26
AKU-375: Logging service updates to filter pubSub logging

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-385 to ensure that pubSub logging is also filtered along with the other log requests.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/Core.js
+++ b/aikau/src/main/resources/alfresco/core/Core.js
@@ -408,11 +408,15 @@ define(["dojo/_base/declare",
             topics = [topics];
          }
 
+         /*jshint -W059 */
+         var callerName = arguments.callee.caller.name;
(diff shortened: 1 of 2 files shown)


aikau/src/main/resources/alfresco/core/Core.js
```diff
@@ -408,11 +408,15 @@ define(["dojo/_base/declare",
             topics = [topics];
          }
 
+         /*jshint -W059 */
+         var callerName = arguments.callee.caller.name;
```
MartinDoyleUK: This (`arguments.callee`) has been removed from JavaScript. Browsers still support it, but for how long? Is there really no alternative?

draperd: I'm aware of this... when it finally stops being supported, we'll do something about it.

PR #345, opened 2015-06-26
AKU-380: Resolved DND nested drag over/out issue

draperd:
This PR finally (hopefully) resolves https://issues.alfresco.com/jira/browse/AKU-380. The issue was that we were implementing the dojo/dnd/Source extension points "onDraggingOut" and "onDraggingOver"... however those extension points were only being fired if the targetState was NOT disabled, e.g.

```JAVASCRIPT
if(this.isDragging && this.targetState != "Disabled"){
  this.onDraggingOut();
}
```

This meant that the updates we'd added to notify the dojo/dnd/Manager of a change of current target were only being fired when the current target was NOT disabled. This meant that as soon as a target was marked as disabled (e.g. when passing over a nested drop target that accepted a different type) the new target was not set when leaving that nested target.

The solution is to use an aspect to chain to the "onOutEvent" and "onOverEvent" to handle all drag out and over events.

There are no tests for this because unfortunately we're not able to get the unit tests to handle the nested targets properly at the moment.

### DIFF
--- a/aikau/src/main/resources/alfresco/dnd/DragAndDropItems.js
+++ b/aikau/src/main/resources/alfresco/dnd/DragAndDropItems.js
@@ -161,7 +161,7 @@ define(["dojo/_base/declare",
          this.sourceTarget = new Source(this.paletteNode, {
             copyOnly: !this.useItemsOnce,
             selfCopy: false,
-            accept: this.acceptTypes || [],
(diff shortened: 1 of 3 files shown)


kevinroast:
excellent

kevinroast:
from my testing this has fixed the problem! working much much better now thanks

PR #346, opened 2015-06-29
AKU-337: Remove Share dependencies from Dashlet

MartinDoyleUK:
This addresses issue [AKU-337](https://issues.alfresco.com/jira/browse/AKU-337) which is about removing the dependency on Share from the Dashlet widget.


aikau/src/main/resources/alfresco/dashlets/Dashlet.js
```diff
@@ -18,263 +18,368 @@
...
-             */
-            titleConfig: null,
+         /**
+          * The container that holds the title.
+          * Will be populated by dojo.
+          *
+          * @instance
+          * @type {HTMLElement}
+          */
+         titleNode: null,
```
draperd: I'm not sure we need to define all of these nodes that are declared in the HTML template, we don't do it anywhere else and it doesn't add a huge amount of value... and it makes the source file that much longer. I appreciate you're just leaving what was there before, but it might be worth removing them.

draperd:
I've merged, but we should still remove those superfluous node declarations

PR #350, opened 2015-06-29
AKU-395: Code updates and test page for infinite scroll in dashlet

draperd:
This PR addresses (in part) https://issues.alfresco.com/jira/browse/AKU-395. I've attempted to reproduce the issues that the customer was having with infinite scroll inside a dashlet (now possible with the pure Aikau dashlet). I've had to make some minor tweaks to get it working and I've added a test page, but no test yet... this might be worth pulling into 1.0.24 though and completing the test work in the next sprint (alternatively I'll try and get the test in before end of sprint)

### DIFF
--- a/aikau/src/main/resources/alfresco/core/_EventsMixin.js
+++ b/aikau/src/main/resources/alfresco/core/_EventsMixin.js
@@ -82,7 +82,7 @@ define(["alfresco/core/Core",
           *                        [functionUtils debounce method]{@link module:alfresco/util/functionUtils#debounce}
           * @return {Object} An object containing a remove() function which will clear any outstanding publish
           */
-          debouncedPublish: function alfresco_core_Core__debouncedPublish(topic, payload, global, parentScope, args) {
+          debouncedPublish: function alfresco_core__EventsMixin__debouncedPublish(topic, payload, global, parentScope, args) {
             var publishFunc = lang.hitch(this, this.alfPublish, topic, payload, global, parentScope),
                debounceArgs = lang.mixin({
                   name: topic,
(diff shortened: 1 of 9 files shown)


aikau/src/main/resources/alfresco/core/_EventsMixin.js
```diff
@@ -82,7 +82,7 @@ define(["alfresco/core/Core",
           *                        [functionUtils debounce method]{@link module:alfresco/util/functionUtils#debounce}
           * @return {Object} An object containing a remove() function which will clear any outstanding publish
           */
-          debouncedPublish: function alfresco_core_Core__debouncedPublish(topic, payload, global, parentScope, args) {
+          debouncedPublish: function alfresco_core__EventsMixin__debouncedPublish(topic, payload, global, parentScope, args) {
```
MartinDoyleUK: Oops!

aikau/src/main/resources/alfresco/dashlets/css/Dashlet.css
```diff
@@ -115,7 +114,12 @@
             line-height: 1.5;
          }
       }
+
+      .alfresco-dashlets-Dashlet__body__widgets {
+         overflow: auto;
+      }
```
MartinDoyleUK: This should be written as below, in order to use the built-in LESS BEM-support.

```
&__widgets {
   overflow: auto;
}
```

draperd: I tried it like that and it didn't work... *where* should you add that?

MartinDoyleUK: Exactly the same place

PR #355, opened 2015-07-01
AKU-377: InfiniteScrollService issue with large Dashlet

MartinDoyleUK:
This addresses issue [AKU-377](https://issues.alfresco.com/jira/browse/AKU-377) and will now fire a resize event which is then re-transmitted as a scroll event to the scroll area, forcing it to check whether it needs to make a new download.


aikau/src/test/resources/alfresco/dashlets/InfiniteScrollDashletTest.js
```diff
@@ -42,61 +42,51 @@ define(["alfresco/TestCommon",
...
-            .pressKeys(keys.ARROW_DOWN)
-            .pressKeys(keys.ARROW_DOWN)
-            .pressKeys(keys.ARROW_DOWN)
-            .pressKeys(keys.ARROW_DOWN)
-            .pressKeys(keys.ARROW_DOWN)
-            .pressKeys(keys.ARROW_DOWN)
-         .end()
-         .getLastPublish("BELOW_ALF_EVENTS_SCROLL")
+         return browser.execute(function() {
+               document.querySelector("#BELOW_DASHLET .alfresco-dashlets-Dashlet__body__widgets").scrollTop += 20;
```
draperd: I know you said that there was a bug with the keyboard navigation, but I'd like to keep the use of the keyboard navigation to ensure that it triggers the reload and that it doesn't reset the focused row back to the first item again.

PR #360, opened 2015-07-02
AKU-378: Only show validation when forms lose focus

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-378 to add the ability to configure forms to not show validation errors immediately. This is done by configuring the form with the new attribute "showValidationErrorsImmediately" as false (the default is true).

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
@@ -1439,6 +1439,38 @@ define(["dojo/_base/declare",
             this.alfLog("warn", "No watch method found on wrapped widget", this);
          }
       },
+
+      /**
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -1439,6 +1439,38 @@ define(["dojo/_base/declare",
...
       },
+
+      /**
+       * This is used to log whether or not this form control has had user focus or not. The reason being that in 
+       * certain circumstances we might not want to show validation errors until the user has at least made an attempt
+       * to enter a value before they are shown validation error messages. 
+       * 
+       * @instance
+       * @type {boolean}
+       * @default false
```
MartinDoyleUK: Not necessary to specify the default for simple types ... just put `@default` and it'll auto-complete in the generated docs

PR #363, opened 2015-07-06
AKU-398: Add a Site Picker

MartinDoyleUK:
This addresses issue [AKU-398](https://issues.alfresco.com/jira/browse/AKU-398) which requests the creation of a dedicated SitePicker control.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
@@ -153,7 +153,7 @@ define(["alfresco/forms/controls/Picker",
                                  {
                                     name: "alfresco/pickers/SitePicker",
                                     config: {
-                                       singleItemMode: this.singleItemModel
+                                       singleItemMode: this.singleItemMode
                                     }
                                  }
                               ],
--- a/aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
@@ -34,7 +34,7 @@
  *                name: "site",
  *                description: "Pick a site, any site",
  *                label: "Site",
- *                multiSite: false
+ *                singleItemMode: true
  *             }
  *          }
  *       ]
@@ -78,7 +78,7 @@ define(["alfresco/forms/controls/Picker",
        * @type {boolean}
        * @default
        */
-      multiSite: false,
+      singleItemMode: true,
 
       /**
        * After config has been mixed in, setup the widget model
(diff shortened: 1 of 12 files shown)


aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
```diff
@@ -0,0 +1,200 @@
...
+   return declare([Picker], {
+
+      /**
+       * An array of the i18n files to use with this widget.
+       *
+       * @instance
+       * @type {Object[]}
+       */
+      i18nRequirements: [
+         {i18nFile: "./i18n/Picker.properties"},
```
draperd: This isn't required... i18n properties will be inherited, there is no need to explicitly add files that the parent widget declares.

aikau/src/main/resources/alfresco/pickers/SitePicker.js
```diff
@@ -0,0 +1,245 @@
...
+      ],
+
+      /**
+       * Whether to select multiple sites in the picker or just one
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       */
+      multiSite: false,
```
draperd: Can we switch this to be "singleItemMode" to make it consistent with other pickers please

aikau/src/main/resources/alfresco/pickers/SitePicker.js
```diff
@@ -0,0 +1,245 @@
...
+ * @author Dave Draper
+ */
+define(["dojo/_base/declare",
+        "alfresco/pickers/Picker"],
+        function(declare, Picker) {
+
+   var sitesView = {
+      name: "alfresco/lists/views/AlfListView",
+      config: {
+         widgets: [
```
draperd: We probably also want to override the "noItemsMessage" message here to be site specific

aikau/src/main/resources/alfresco/pickers/SitePicker.js
```diff
@@ -0,0 +1,245 @@
...
+                        label: "picker.allSites.label",
+                        publishTopic: "ALF_ADD_PICKER",
+                        publishPayload: {
+                           currentPickerDepth: 0,
+                           pickerLabel: "AllSites",
+                           picker: [
+                              {
+                                 name: "alfresco/lists/AlfSortablePaginatedList",
+                                 config: {
+                                    itemsProperty: "",
```
draperd: Although the view will have a "noItemsMessage" we might also want to look at the other messages that are displayed by a list as they're document centric by default.

aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
MartinDoyleUK: I assumed that overwriting the property in a subclass would remove any parent files, like a normal overridden property. Is that not the case then?

draperd: No, it doesn't work like that... the i18nRequirements and the cssRequirements are only used when Surf analyses the source of a widget to work out its dependencies (this is why you can't configure in additional CSS and i18n dependencies are in a WebScript controller). So because the parent Picker is recognized as a dependency, its i18n dependencies will be included automatically.

aikau/src/main/resources/alfresco/pickers/SitePicker.js
MartinDoyleUK: This is configured elsewhere

PR #365, opened 2015-07-06
AKU-390: Nested DND re-order and delete updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-390 to resolve issues with the DND capabilities in nested items. Previous tests for re-ordering were only looking at root level items so a new test page has been created with items nested within nested items and these are then re-ordered (using the up/down buttons), moved from one target to another and deleted. 

Some significant portions of the DND implementation have required updating to ensure that this now works as it should.

### DIFF
--- a/aikau/src/main/resources/alfresco/dnd/DroppedNestingItemWrapper.js
+++ b/aikau/src/main/resources/alfresco/dnd/DroppedNestingItemWrapper.js
@@ -29,12 +29,14 @@
 define(["dojo/_base/declare",
         "alfresco/dnd/DroppedItemWrapper",
         "alfresco/dnd/Constants",
+        "alfresco/core/ObjectTypeUtils",
         "dojo/_base/lang",
         "dojo/_base/array",
         "dojo/on",
         "dojo/Deferred",
-        "dijit/registry"], 
-        function(declare, DroppedItemWrapper, Constants, lang, array, on, Deferred, registry) {
+        "dijit/registry",
+        "jquery"], 
+        function(declare, DroppedItemWrapper, Constants, ObjectTypeUtils, lang, array, on, Deferred, registry, $) {
    
    return declare([DroppedItemWrapper], {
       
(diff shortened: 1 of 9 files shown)


aikau/src/main/resources/alfresco/dnd/DroppedNestingItemWrapper.js
```diff
@@ -114,34 +133,142 @@ define(["dojo/_base/declare",
...
+            }
          }
       },
 
       /**
+       * This function can be called to update wrapping widgets that the current item has changed.
+       *
+       * @instance
+       */
+      onUpdateParentOfChange: function alfresco_dnd_DroppedNestingItemWrapper__onUpdateParentOfChange() {
```
MartinDoyleUK: This is not an ideally-named method. The "on" prefix suggests that an event has just happened and this is the handler, but in fact this is the method that triggers the event. A more suitable name might be "updateParentOfChange", or even more accurately "notifyParentOfChange".

aikau/src/test/resources/alfresco/dnd/NestedReorderTest.js
```diff
@@ -0,0 +1,150 @@
...
+         return browser.findByCssSelector(".middleTestItemWrapper:nth-child(1) .innerTestItemWrapper:nth-child(1) .down")
+            .click()
+         .end()
+         .clearLog()
+         .findByCssSelector(".confirmationButton > span")
+            .click()
+         .end()
+         .getLastPublish("FORM1_POST")
+            .then(function(payload) {
+               assert.deepPropertyVal(payload, "data.0.config.widgets.0.config.widgets.0.name", "E", "E was not saved as first item");
```
MartinDoyleUK: Square-bracket notation does work in this assert function, so `data[0].config.widgets[0].config.widgets[0].name` would be equally valid, and more usual for addressing the indices of an array.

PR #373, opened 2015-07-09
AKU-403: Dashlet theming

MartinDoyleUK:
This addresses issue [AKU-403](https://issues.alfresco.com/jira/browse/AKU-403), which requires that Dashlets can be themed in Share, in order to sit alongside the non-Aikau dashlets when the theme is changed from the default. See the JIRA ticket for sample files of how to theme them in Share.

### DIFF
--- a/aikau/src/main/resources/alfresco/css/less/defaults.less
+++ b/aikau/src/main/resources/alfresco/css/less/defaults.less
@@ -16,7 +16,7 @@
 @header-dropdown-menu-font-color: #fff;
 @header-hover-background-color: #717171;
 @header-hover-font-color: #fff;
-@header-focus-background-color: #4D4D4D;
+@header-focus-background-color: #4d4d4d;
 @header-focus-font-color: #fff;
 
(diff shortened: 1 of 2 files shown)


aikau/src/main/resources/alfresco/css/less/defaults.less
```diff
@@ -50,6 +50,19 @@
 @de-emphasized-font-color: #999;
 @highlighted-font-color: #fff;
 
+// Dashlets
+@dashlet-background: #fff;
+@dashlet-border: 1px solid #ccc;
```
draperd: Just a minor thing... but I was thinking that where we have a "global" variable, e.g. @standard-border we should reference it in the more specific examples, e.g 

```
@dashlet-border: @standard-border
```

That way borders can be universally changed, or you could just override borders for dashlets. Anopther example would be:

@dashlet-background = @primary-background-color

I'm not sure if there are any others, we could look to add others, like @standard-border-radius?

PR #376, opened 2015-07-10
AKU-267: Repeating form dialog support in DialogService

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-267 to add support for repeating form dialogs (the create and create another paradigm). Unit tests have been updated accordingly. Both golden path and error handling scenarios have been addressed.


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -158,12 +164,107 @@ define(["dojo/_base/declare",
...
+      _formCancellationTopic: "ALF_CLOSE_DIALOG",
+
+      /**
+       * The default configuration for form dialogs. This is used as a base when requests are received.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       */
+      defaultFormDialogConfig: {
```
MartinDoyleUK: This isn't a string, and the @default without a value won't work with complex objects

PR #383, opened 2015-07-15
Update ContentService test to add Upload button

MartinDoyleUK:
Update ContentService test to add Upload button as well as converting ContentService test page to use DebugLog and make certain selectors less fragile.

### DIFF
--- a/aikau/src/test/resources/alfresco/services/ContentServiceTest.js
+++ b/aikau/src/test/resources/alfresco/services/ContentServiceTest.js
@@ -22,9 +22,9 @@
  * @author Dave Draper
  */
 define(["intern!object",
-        "intern/chai!assert",
-        "require",
-        "alfresco/TestCommon"],
+        "intern/chai!assert", 
+        "require", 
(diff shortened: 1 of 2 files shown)


aikau/src/test/resources/alfresco/services/ContentServiceTest.js
```diff
@@ -22,9 +22,9 @@
  * @author Dave Draper
  */
 define(["intern!object",
-        "intern/chai!assert",
-        "require",
-        "alfresco/TestCommon"],
+        "intern/chai!assert", 
+        "require", 
+        "alfresco/TestCommon"],
```
davidcognite: Trailing whitespace? We should decide whether to striping it or not (my IDE does, it seems yours doesn't).

MartinDoyleUK: I think stripping is fine. My editor seems to have lost its setting for doing so (now restored). Do you need me to push an update with the whitespace gone, or are you happy to merge as is?

PR #384, opened 2015-07-15
AKU-423: DialogService form to allow publish scoping

MartinDoyleUK:
This addresses issue [AKU-423](https://issues.alfresco.com/jira/browse/AKU-423) which requests allowing a custom scope to be passed to DialogService for use when submitting a form. The publishPayload property will now the following config parameters: formSubmissionTopic [existing], formSubmissionPayloadMixin [existing], formSubmissionGlobal and formSubmissionScope. Core.js has also been updated to allow proper use of a custom, one-off scope which will also update the alfPublishScope property in the payload. Tests have all been updated accordingly.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/Core.js
+++ b/aikau/src/main/resources/alfresco/core/Core.js
@@ -399,8 +399,9 @@ define(["dojo/_base/declare",
        * @param {object} payload The payload to publish on the supplied topic
        * @param {boolean} [global] Indicates that the pub/sub scope should not be applied
        * @param {boolean} [parentScope] Indicates that the pub/sub scope inherited from the parent should be applied
+       * @param {String} [customScope] An optional scope to use for this topic-publish
        */
-      alfPublish: function alfresco_core_Core__alfPublish(topics, payload, global, parentScope) {
+      alfPublish: function alfresco_core_Core__alfPublish(topics, payload, global, parentScope, customScope) {
 
          // Allow an array of topics so we can publish multiple ones.
          if (!lang.isArray(topics))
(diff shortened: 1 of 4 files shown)


aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/services/DialogService.get.js
```diff
@@ -46,6 +46,33 @@ model.jsonModel = {
...
+         config: {
+            label: "Create Form Dialog (with custom scoping)",
+            publishTopic: "ALF_CREATE_FORM_DIALOG_REQUEST",
+            publishPayload: {
+               dialogId: "SCOPED_FORM",
+               dialogTitle: "Scoped Form Dialog",
+               formSubmissionTopic: "CUSTOM_FORM_TOPIC",
+               formSubmissionGlobal: false,
+               formSubmissionToParent: false,
+               formSubmissionScope: "CUSTOM_FORM_SCOPE_",
```
davidcognite: Is it worth testing the formSubmissionToParent does what it should?

aikau/src/main/resources/alfresco/core/Core.js
```diff
@@ -399,8 +399,9 @@ define(["dojo/_base/declare",
        * @param {object} payload The payload to publish on the supplied topic
        * @param {boolean} [global] Indicates that the pub/sub scope should not be applied
        * @param {boolean} [parentScope] Indicates that the pub/sub scope inherited from the parent should be applied
+       * @param {String} [customScope] An optional scope to use for this topic-publish
        */
```
davidcognite: Worth documenting that customScope requires global and parentScope to both be false?

PR #385, opened 2015-07-16
AKU-431: NumberSpinner cannot be empty

MartinDoyleUK:
This addresses issue [AKU-431](https://issues.alfresco.com/jira/browse/AKU-431) by allowing a new config option of 'permitEmpty' on the NumberSpinner control and update tests accordingly.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
@@ -46,7 +46,7 @@ define(["alfresco/forms/controls/BaseFormControl",
        *
        * @instance
        * @type {number}
-       * @default 0
+       * @default
        */
       min: null,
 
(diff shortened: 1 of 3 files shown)


aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
```diff
@@ -109,7 +120,7 @@ define(["alfresco/forms/controls/BaseFormControl",
             // See AKU-341 for details of why we handle commas and spaces...
             var value = this.wrappedWidget.textbox.value.replace(",","").replace(" ","");
             var parsedValue = parseFloat(value);
-            var isValid = !isNaN(parsedValue);
+            var isValid = !isNaN(parsedValue) || (lang.trim(value) === "" && this.permitEmpty);
```
davidcognite: Very minor: would this be more performant if the options were reversed in the second part of the OR: (this.permitEmpty && lang.trim(value) === "")

aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
```diff
@@ -55,11 +55,22 @@ define(["alfresco/forms/controls/BaseFormControl",
        *
        * @instance
        * @type {number}
-       * @default null
+       * @default
```
davidcognite: I've just checked JSDoc docs and yes, you're right this is much better patten; we don't need to specify the default value - good spot.

PR #387, opened 2015-07-16
AKU-425: Resolve Sonar Blocker and Critical Issues

MartinDoyleUK:
This addresses issue [AKU-425](https://issues.alfresco.com/jira/browse/AKU-425), which is about resolving all of the highest-priority issues on Sonar. This has been done by fixing JSHint bugs in all files [highlighted by Sonar] and additionally fixing specific bugs mentioned but not picked up by JSHint.

A full test-suite was run successfully before submitting this pull-request.

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/AlfDocumentFilters.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/AlfDocumentFilters.js
@@ -36,10 +36,9 @@ define(["dojo/_base/declare",
         "dojo/_base/lang",
         "dojo/_base/array",
         "dojo/dom-construct",
-        "dojo/dom-class",
-        "dojo/on"],
+        "dojo/dom-class"],
         function(declare, Twister, _OnDijitClickMixin, _AlfDocumentListTopicMixin, template,
-                 AlfDocumentFilter, lang, array, domConstruct, domClass, on) {
+                 AlfDocumentFilter, lang, array, domConstruct, domClass) {
    return declare([Twister, _OnDijitClickMixin, _AlfDocumentListTopicMixin], {
 
       /**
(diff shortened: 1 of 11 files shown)


aikau/src/main/resources/alfresco/documentlibrary/AlfDocumentFilters.js
```diff
@@ -154,11 +153,11 @@ define(["dojo/_base/declare",
       /**
        * @instance
        */
-      onShowMoreClick: function alfresco_documentlibrary_AlfDocumentFilters__onShowMoreClick(evt) {
+      onShowMoreClick: function alfresco_documentlibrary_AlfDocumentFilters__onShowMoreClick() {
```
davidcognite: This feels inconsistent. Why remove the unused param here, but annotate others with "unused:false"?

aikau/src/main/resources/alfresco/preview/PdfJs/DocumentView.js
```diff
@@ -309,10 +310,10 @@ define(["dojo/_base/declare",
...
             var currentPage = this.pages[0],
                 container = currentPage.container,
-                hmargin = parseInt(domStyle.get(container, "margin-left")) + parseInt(domStyle.get(container, "margin-right")),
-                vmargin = parseInt(domStyle.get(container, "margin-top")),
-                contentWidth = parseInt(currentPage.content.pageInfo.view[2]),
-                contentHeight = parseInt(currentPage.content.pageInfo.view[3]),
+                hmargin = parseInt(domStyle.get(container, "margin-left"), 10) + parseInt(domStyle.get(container, "margin-right"), 10),
+                vmargin = parseInt(domStyle.get(container, "margin-top"), 10),
+                contentWidth = parseInt(currentPage.content.pageInfo.view[2], 10),
+                contentHeight = parseInt(currentPage.content.pageInfo.view[3], 10),
```
davidcognite: Is it worth updating the .jshintrc file to pick up this (and other?) errors prior to commit, rather than wait for a sonar run in future?

aikau/src/main/resources/alfresco/documentlibrary/AlfDocumentFilters.js
MartinDoyleUK: When it's an inherited method, I annotate and when it's only used in the instance, I remove it as redundant. At least that was the intention.

aikau/src/main/resources/alfresco/preview/PdfJs/DocumentView.js
MartinDoyleUK: The decision when I introduced JSHint was that we didn't want to enforce JSHint yet, only have it in an advisory role. It might be worth reconsidering, but there's still the problem that there are over 1,000 JSHint warnings that we would need to cleanup first. Once we were in that position though, we could certainly put JSHint into Bamboo and run that on commit.

PR #388, opened 2015-07-16
AKU-436: Notifications show off screen if scrolled

MartinDoyleUK:
This addresses issue [AKU-436](https://issues.alfresco.com/jira/browse/AKU-436) which fixes the notifications showing off screen when scrolled by changing positioning from absolute to fixed. Have also added a requested close button on the notification.

### DIFF
--- a/aikau/src/main/resources/alfresco/notifications/css/AlfNotification.css
+++ b/aikau/src/main/resources/alfresco/notifications/css/AlfNotification.css
@@ -5,7 +5,7 @@
    position: fixed;
    right: 0;
    top: 52.5%;
-   transition: opacity .2s ease-out, top .2s ease-out;
+   transition: all .2s ease-out;
    z-index: 9999; /* This isn't ideal, but searching all z-indexes is even worse (better solution possible?) */
    &__container {
       background: #666;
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/notifications/css/AlfNotification.css
```diff
@@ -13,11 +13,23 @@
       box-shadow: @standard-box-shadow;
       box-sizing: border-box;
       margin: 0 auto;
-      padding: 20px;
+      padding: 20px 30px 20px 20px;
```
davidcognite: Could these use LESS vars rather than hardcoded px value?

aikau/src/main/resources/alfresco/notifications/css/AlfNotification.css
```diff
@@ -13,11 +13,23 @@
...
       position: relative;
       top: -50%;
       width: 400px;
    }
+   &__close {
+      color: #aaa;
+      cursor: pointer;
+      font-size: 25px;
```
davidcognite: LESS var?

aikau/src/main/resources/alfresco/notifications/templates/AlfNotification.html
```diff
@@ -1,5 +1,6 @@
 <div class="alfresco-notifications-AlfNotification">
    <div class="alfresco-notifications-AlfNotification__container">
+      <div class="alfresco-notifications-AlfNotification__close" data-dojo-attach-event="click:_onCloseClick">&times;</div>
```
davidcognite: Is this a pattern we should be using (embedding behaviour in markup)?

MartinDoyleUK: This is BEM, which is definitely intended. We use double-underscore to delineate the OO model of the code and double-hyphen to indicate states. It should not be embedding behaviour, but instead annotating by state and item and leaving the CSS to determine how it looks and the JS to determine what it does.

davidcognite: For the record, we discussed this offline and I don't like the pattern (personal preference), but there's nothing technically wrong with it and it's consistent with Aikau patterns used elsewhere.

PR #394, opened 2015-07-22
AKU-428: Fix issues with Gallery view not resizing correctly.

davidcognite:


aikau/src/main/resources/alfresco/lists/views/layouts/Grid.js
```diff
@@ -257,19 +257,22 @@ define(["dojo/_base/declare",
...
          domStyle.set(node, {"width": widthToSet});
          var dimensions = {
-            w: widthToSet,
-            h: null
-         };
-         array.forEach(node.children, lang.hitch(this, "resizeWidget", dimensions));
+               w: widthToSet,
+               h: null
+            },
+            widgetsToResize = query("[widgetId]", node); // Resize all contained widgets, not just immediate children.
```
MartinDoyleUK: Are we happy calling resize on every descendant widget, given that Dojo widgets often resize their children as well, when their own resize function is called? Is this definitely better than just finding the first descendant-generation with  resize functions on them (although obviously it's much more succinct)?

davidcognite: I discussed this with Martin this morning and we felt that the additional code required to resize only the first child with a resize method, along with the need for every widget that contains other widgets to pass on a resize method it receives outweighed the risks of some widgets receiving two resize events.

PR #397, opened 2015-07-23
AKU-459: Ability to Set a New User Default Page Preference Via Publish

rgauss:
- Added _UserServiceTopicMixin which contains all UserService topics
  - Updated UserService to use topics from mixing
  - Updated UserService with onSetUserDefaultPage method and related
success and failure methods
  - Added check for optional alfResponseTopic in
PreferenceService.setPreference
  - Added UserServiceDefaultPage test webscript
  - Added User Service Default Page Test suite to UserServiceTest

### DIFF
--- a/aikau/src/main/resources/alfresco/services/UserService.js
+++ b/aikau/src/main/resources/alfresco/services/UserService.js
@@ -22,19 +22,33 @@
  * @extends module:alfresco/core/Core
  * @mixes module:alfresco/core/CoreXhr
  * @mixes module:alfresco/core/NotificationUtils
+ * @mixes module:alfresco/services/_UserServiceTopicMixin
+@mixes module:alfresco/services/_PreferenceServiceTopicMixin
  * @author Dave Draper
  */
 define(["dojo/_base/declare",
         "alfresco/core/Core",
         "alfresco/core/CoreXhr",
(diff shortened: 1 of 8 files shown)


aikau/src/main/resources/alfresco/services/UserService.js
```diff
@@ -22,19 +22,33 @@
  * @extends module:alfresco/core/Core
  * @mixes module:alfresco/core/CoreXhr
  * @mixes module:alfresco/core/NotificationUtils
+ * @mixes module:alfresco/services/_UserServiceTopicMixin
+@mixes module:alfresco/services/_PreferenceServiceTopicMixin
```
MartinDoyleUK: Missing an asterisk at start of line

MartinDoyleUK:
Apart from the above syntactic comment, the rest looks good. As it's an area of the codebase I don't know though, I think we'll wait until @draperd is back on Monday to review and merge this pull-request.

rgauss:
Fixed the missing asterisk, thanks.

davidcognite:
I've had a look, and this seems fairly straight forward addition to the service, so I'm happy to merge this. I like the topic extraction into a separate file, that dovetails nicely with thoughts Martin and I have had.

PR #410, opened 2015-07-30
NavigationService: allow queryParams and hashParams objects in payload to support encoding

davidcognite:
This is required to support RM-2465.


MartinDoyleUK:
Needs to handle addQueryParam when hash already present in URL

davidcognite:
Support for relative URLs now added. :-)

MartinDoyleUK:
Why do we need the absolute URL handling section at all? It looks like the string-parsing fallback for relative URLs would also work for absolute URLs, so isn't it just adding complexity and length to the source file that's not necessary?

davidcognite:
It's a shame, as that code is cleaner (which is why I wanted to keep it), but you're right. Lets remove code that isn't useful. I've kept the parseUrl method, though, as I think that's a useful util to have.

PR #413, opened 2015-07-30
Add dedicated CSS resource for NumberSpinner control

kevinroast:
Styles to remove the default dijit transitions, gradients and border colours that we don't use in other Alfresco Form controls.


draperd:
I'm going to leave this to Martin to merge as he's working on a bug relating to the NumberSpinner, it it looks fine to me though.

PR #414, opened 2015-07-30
AKU-405: Added support for form controls in tabs

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-405 to add a new mixin module "alfresco/forms/LayoutMixin" that can be applied to layout controls to create layout widgets to be used inside forms. This has been abstracted and applied to the "alfresco/forms/ControlRow" widget and used to create the "alfresco/forms/TabbedControls" and "alfresco/forms/ControlColumn" modules.

Unit tests have been added to support the new modules.

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
+++ b/aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
@@ -270,6 +270,18 @@ define(["dojo/_base/declare",
       tabDeletionTopic: null,
 
       /**
+       * This is the default delayed processing behaviour of tabs. By default a tabs content
+       * will not be created until it is displayed (unless the tab is configured to request
+       * otherwise). However, this can be overridden to switch the default so that all tab
+       * content will be created as soon as the container is created.
+       *
+       * @instance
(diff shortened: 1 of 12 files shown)


aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
```diff
@@ -290,12 +306,24 @@ define(["dojo/_base/declare",
             // we'll ensure that the first tab is both selected and will be immediately rendered
             var tabSelected = false;
             array.forEach(this.widgets, function(widget) {
-               if (widget.delayProcessing !== false && !widget.selected)
+               if ((widget.delayProcessing === true && !widget.selected) || 
+                   widget.delayProcessing === false)
```
MartinDoyleUK: Could perhaps be rewritten as `if (!widget.delayProcessing || !widget.selected) {`

draperd: What happens if it's marked as false and isn't marked as selected? The conditional blcok wouldn't be entered

PR #432, opened 2015-08-03
AKU-449: Dialog loses focus and escape doesn't close

MartinDoyleUK:
This addresses issue [AKU-449](https://issues.alfresco.com/jira/browse/AKU-449) where a dialog can lose focus and escape will no longer close the dialog. A stack of active dialogs is now maintained by the DialogService and pressing escape will always close the dialog.


draperd:
No unit test?

MartinDoyleUK:
Unable to create repeatable Leadfoot/Intern test steps, so just had to use manual verification.

PR #435, opened 2015-08-05
AKU-472: Search without URL hashing

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-472, https://issues.alfresco.com/jira/browse/AKU-475 and https://issues.alfresco.com/jira/browse/AKU-488 to make a number of changes required to enable the search widgets to be used with useHash configured to be false.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfSortablePaginatedList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfSortablePaginatedList.js
@@ -240,7 +240,7 @@ define(["dojo/_base/declare",
                }
                else
                {
-                  this.loadData();
+                  this.onReloadData();
                }
             }
          }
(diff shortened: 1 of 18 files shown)


aikau/src/main/resources/alfresco/lists/AlfSortablePaginatedList.js
```diff
@@ -240,7 +240,7 @@ define(["dojo/_base/declare",
                }
                else
                {
-                  this.loadData();
+                  this.onReloadData();
```
MartinDoyleUK: Is this reacting to a reloadData event, or requesting a data-reload? Just querying the new prefix.

draperd: The onReloadData function was added in a previous release and the AlfSearchList is overriding it. It's typically called as a result of publishing the reload data topic, however we need to call it directly here to get the overridden behaviour in AlfSearchList when sorting.

PR #441, opened 2015-08-07
AKU-500: Update response scope behaviour to handle empty string falsy logic

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-500 and re-works the logic for figuring out the correct alfResponseScope to apply to a payload to ensure that global scope (or empty string) is accepted as a valid scope. I've added some specific unit tests to verify the expected behaviour and checked that the issue no longer occurs on Share with the Document Library.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/DialogService.js
+++ b/aikau/src/main/resources/alfresco/services/DialogService.js
@@ -743,22 +743,11 @@ define(["dojo/_base/declare",
             }
 
             // Mixin in any additional payload information...
-            // Use the alfResponseScope from the current payload if a responseScope isn't available,
-            // but values in the formSubmissionPayloadMixin should trump both...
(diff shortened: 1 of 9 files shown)


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -744,19 +743,37 @@ define(["dojo/_base/declare",
...
+               lang.mixin(data, {
+                  responseScope: payload.responseScope
+               });
+            }
+            else if (payload.alfResponseScope || payload.alfResponseScope === "")
+            {
+               lang.mixin(data, {
+                  responseScope: payload.alfResponseScope
+               });
+            }
```
MartinDoyleUK: As discussed, could replace lines 750-761 with just:

```javascript
lang.mixin(data, {
  responseScope: payload.alfResponseScope
});
```

PR #445, opened 2015-08-07
AKU-452: Update Aikau buttons to match new Share improvements

MartinDoyleUK:
This addresses issue [AKU-452](https://issues.alfresco.com/jira/browse/AKU-452), which is about making updates to the Aikau buttons to match the new Share improvements.

The final effect should be invisible to end-users of Share, however the "Aikau Test Theme"—as applied to our test application—will make the new changes visible on all unit-test pages.

NOTE: If at all possible, please double-check the aesthetics for me (both old and new) to ensure I've not broken the old and that it matches the new.

### DIFF
--- a/aikau/src/main/resources/alfresco/buttons/css/AlfButton.css
+++ b/aikau/src/main/resources/alfresco/buttons/css/AlfButton.css
@@ -2,10 +2,14 @@
    background: @standard-button-background;
    border: @standard-button-border;
    border-radius: @standard-button-border-radius;
+   box-sizing: border-box;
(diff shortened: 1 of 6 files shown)


draperd:
As discussed we need to make some changes here.. the Aikau test app theme should just be setting colours, the defaults.less should have the new styling but with the old colours. Then it's just a case of the 5.1 themes setting the new colours (we possibly want to consider font-weight, border thickness, radius etc) but ultimately the theme file should be very simple in what is getting defined.

PR #449, opened 2015-08-10
AKU-489: AlfTabContainer and AlfSideBarContainer heights

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-489 and updates the AlfTabContainer so that it creates its tabs in the DOM model before creating widgets in them (this allows the widgets in the tabs to have access to actual dimensions to work from). The AlfSideBarContainer is also updated to make better use of the _ResizeMixin capabilities. The styling of the tabs has been tidied up (to remove the default padding) which has allowed the sidebar borders to be tidied up as well. A unit test has been added to verify that the heights are now calculated sensibly.

### DIFF
--- a/aikau/src/test/resources/alfresco/layout/AlfTabContainerTest.js
+++ b/aikau/src/test/resources/alfresco/layout/AlfTabContainerTest.js
@@ -632,4 +632,39 @@ define(["intern!object",
          TestCommon.alfPostCoverageResults(this, browser);
       }
    });
+
+   registerSuite({
+      name: "Tab Container Tests (height calculations)",
+
+      setup: function() {
+         browser = this.remote;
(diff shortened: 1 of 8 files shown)


aikau/src/test/resources/alfresco/layout/AlfTabContainerTest.js
```diff
@@ -632,4 +632,39 @@ define(["intern!object",
...
+            .getSize()
+            .then(function(size) {
+               height = size.height;
+            })
+         .end()
+         .findById("SIDEBAR")
+            .getSize()
+               .then(function(size) {
+                  var target = height - 46;
+                  assert(size.height > (target - 5) && size.height < (target + 5), "The sidebar height (" + size.height + ") was not within a margin of error of the expected height (" + target + ")");
```
MartinDoyleUK: Would be more ideal to use `assert.closeTo(size.height, target, 5, "Sidebar height incorrect");`

PR #453, opened 2015-08-12
AKU-492: Improvements in JS logic to support DocLib options

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-492 to improve the logic handling relating to siteId, containerId and rootNode values that might be set via calls to the functions provided in the doclib.lib.js. Previously any undefined values were being interpreted as actual data because explicit checks for null were being made.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/SearchService.js
+++ b/aikau/src/main/resources/alfresco/services/SearchService.js
@@ -141,51 +141,54 @@ define(["dojo/_base/declare",
        * @param {object} payload The payload defining the document to retrieve the details for.
        */
       onSearchRequest: function alfresco_services_SearchService__onSearchRequest(payload) {
-         if (payload == null || payload.term == null)
+         // jshint maxlen:300, maxcomplexity:false
+         if (!payload || !payload.term)
          {
             this.alfLog("warn", "A request was made to perform a search but no 'term' attribute was provided", payload, this);
          }
          else
          {
             // Check to see whether or not a success topic has been provided...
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/services/SearchService.js
```diff
@@ -196,7 +199,7 @@ define(["dojo/_base/declare",
             }
 
             var sort = "";
-            if (payload.sortField != null && (payload.sortField === "null" || payload.sortField === ""))
+            if (payload.sortField && payload.sortField === "null")
```
MartinDoyleUK: if it has to be equal to the string value `"null"` then why not just `if (payload.sortField === "null")`?

aikau/src/main/resources/alfresco/services/SearchService.js
```diff
@@ -206,25 +209,26 @@ define(["dojo/_base/declare",
             }
             else
             {
-               sort = ((payload.sortField != null) ? payload.sortField : this.sort) + "|" + 
-                      ((payload.sortAscending != null) ? payload.sortAscending : this.sortAscending);
+               sort = ((payload.sortField) ? payload.sortField : this.sort) + "|" + 
+                      ((payload.sortAscending) ? payload.sortAscending : this.sortAscending);
             }
```
MartinDoyleUK: This is a bit difficult to read with so many brackets and such repetition. Would you consider replacing with the below?

```
sort = (payload.sortField || this.sort) + "|" + (payload.sortAscending || this.sortAscending);
```

PR #454, opened 2015-08-12
AKU-519: New Icons for User Home Page and Navigation

rgauss:
- Added icons and CSS


draperd:
Are these icons just for menus? I can't remember the history of that Icons.css file, but I'm wondering if these should go in alfresco/menus/_AlfMenutItemMixin.css and the images in the relative path (e.g. alfresco/menus/css/images). I just want to make sure that we put these in the right place going forwards. It looks like that Icons.css file has just been there since Day 1 of us abstracting Aikau from Share and I can't remember its purpose.

I'd like to confer with you and the team before we merge this - I want to avoid setting any incorrect precedents on these types of PRs

rgauss:
> Are these icons just for menus?

That's what they're designed for, but there's no reason developers couldn't use them elsewhere.

I too was just following precedence of where the other icons in that menu were coming from.

If the location needs to change we can move the others before or after this pull request, but IMHO they should all live in the same place.

draperd:
Icons are currently handled badly for a number of reasons, not least because they're not easily customized so we will want to re-work them in the future anyway.  You're right that this is consistent with the current items in that menu though so I'll merge.

PR #458, opened 2015-08-13
AKU-503: InlineEditProperty updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-503 and has made updates to both the alfresco/renderers/InlineEditProperty widget and how it is used within the various views. It also updates in particular how the cm:title is displayed in views. There is a new check within the InlineEditProperty widget on whether or not the current user has edit permissions (this is a configurable property) and editing will automatically be disabled if they do not. After discussion with UX it was agreed to always display the title and display a warning if the property is not actually set, but we will further review this in the future. Unit tests have been updated.

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/views/AlfDetailedViewItem.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/views/AlfDetailedViewItem.js
@@ -146,16 +146,19 @@ define(["alfresco/lists/views/layouts/Row",
             }
          }, {
             _attachPoint: "title",
-            _render: function(item) {
-               return item.node.properties["cm:title"];
-            },
+            // _render: function(item) {
+            //    return item.node.properties["cm:title"];
+            // },
             id: "DETAILED_VIEW_TITLE",
             name: "alfresco/renderers/InlineEditProperty",
             config: {
                propertyToRender: "node.properties.cm:title",
                postParam: "prop_cm_title",
+               renderSize: "small",
                renderedValuePrefix: "(",
-               renderedValueSuffix: ")"
+               renderedValueSuffix: ")",
+               warnIfNotAvailable: true,
+               warnIfNotAvailableMessage: "no.title.message"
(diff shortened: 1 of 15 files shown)


aikau/src/main/resources/alfresco/documentlibrary/views/AlfDetailedViewItem.js
```diff
@@ -146,16 +146,19 @@ define(["alfresco/lists/views/layouts/Row",
             }
          }, {
             _attachPoint: "title",
-            _render: function(item) {
-               return item.node.properties["cm:title"];
-            },
+            // _render: function(item) {
+            //    return item.node.properties["cm:title"];
+            // },
             id: "DETAILED_VIEW_TITLE",
```
MartinDoyleUK: Can we delete this? It'll be in Git if needed again.

aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
```diff
@@ -207,33 +260,25 @@ define(["dojo/_base/declare",
...
        */
-      hiddenDataRules: null,
+      postCreate: function alfresco_renderers_InlineEditProperty__postCreate() {
+         this.inherited(arguments);
+         if (this.permissionProperty)
+         {
+            var hasEditPermission = lang.getObject(this.permissionProperty, false, this.currentItem);
+            if (!hasEditPermission)
+            {
+               domStyle.set(this.editIconNode, "display", "none");
```
MartinDoyleUK: Can we instead use something like ...

```
domClass.add(this.domNode, "alfresco-renderers-InlineEditProperty--disabled");
```

... instead? Semantically, it's much better and gives greater scope for further changes just by tweaking the CSS, especially when it comes to external customisation. Direct style manipulation is also impossible to override with a stylesheet unless people use `!important` which is obviously __ABadThing<sup>TM</sup>__.

PR #459, opened 2015-08-13
AKU-498: Add configurable focus highlight support

draperd:
This addresses https://issues.alfresco.com/jira/browse/AKU-498


aikau/src/main/resources/alfresco/lists/views/layouts/css/_MultiItemRendererMixin.css
```diff
@@ -2,7 +2,7 @@
    background-color: @list-hover-color;
 }
 
-.alfresco-lists-views-layout-_MultiItemRendererMixin--item:focus {
+.alfresco-lists-views-layout-_MultiItemRendererMixin--item__focusHighlighting:focus {
```
MartinDoyleUK: `--` should be for states (i.e. adjectives) and `__` for objects (i.e. nouns) in the BEM hierarchy naming. I realise it's not a change introduced in this PR, but it would be good to try and make sure we get it right, for consistency and to avoid confusion.

PR #460, opened 2015-08-14
AKU-290: Support for pagination hiding on list loading

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-290 to add support for the alfresco/lists/Paginator to be configured to hide when lists request data and displayed when data is successfully retrieved. The default behaviour is for this to be off because it preserves existing behaviour and also the data request/response topics are variable/configurable. Unit tests have been added to verify behaviour.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/Paginator.js
+++ b/aikau/src/main/resources/alfresco/lists/Paginator.js
@@ -87,32 +87,15 @@ define(["dojo/_base/declare",
       i18nRequirements: [{i18nFile: "./i18n/Paginator.properties"}],
       
       /**
-       * The number of documents to show per page.
-       * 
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/lists/Paginator.js
```diff
@@ -96,15 +115,90 @@ define(["dojo/_base/declare",
...
+       */
+      pageForwardTopic: "ALF_PAGE_FORWARD",
+
+      /**
+       * This will be initialised to be a select menu for the current page.
        * 
        * @instance
-       * @type {number} 
+       * @type {object}
+       * @default null
```
MartinDoyleUK: Just `@default` is sufficient (examples below as well)

PR #461, opened 2015-08-14
AKU-506: Make FixedHeaderFooter resize consistently with other layout widgets

draperd:
This is a follow on to the previous fix for https://issues.alfresco.com/jira/browse/AKU-506 to address some inconsistencies between resizing behaviour of the alfresco/layout/FixedHeaderFooter widget and the other layout widgets. I've tried to make the behaviour more consistent and more expected - one of the main differences here is to switch the default of the "recalculateAutoHeightOnResize" attribute as that was causing confusion when not set with the default "auto" sizing.

This also addresses https://issues.alfresco.com/jira/browse/AKU-483 which wasn't included in the current sprint but made sense to implement whilst I was working on the module and assumes that all numerical heights provided are intended to be pixels (as this also previously caught people out).

I've updated the unit test pages to ensure that these scenarios are covered, the existing tests sufficiently ensure correct rendering.

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/FixedHeaderFooter.js
+++ b/aikau/src/main/resources/alfresco/layout/FixedHeaderFooter.js
@@ -25,16 +25,13 @@
  * widget models.</p>
  * <p>The overall [height]{@link module:alfresco/layout/FixedHeaderFooter#height} of the widget can be explicitly 
  * set but can also be left (or set) to the default value of "auto" which will make the widget take up all the available
- * space in the client window below it. It is also possible to 
- * [configure]{@link module:alfresco/layout/FixedHeaderFooter#recalculateAutoHeightOnResize} the widget so that it will
- * automatically reset the height as the window changes in size.</p>
+ * space in the client window below it.</p>
  *
  * @example <caption>Example configuration:</caption>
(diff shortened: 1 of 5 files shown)


aikau/src/main/resources/alfresco/layout/FixedHeaderFooter.js
```diff
@@ -277,8 +270,8 @@ define(["alfresco/core/ProcessWidgets",
        *
        * @instance
        */
-      _resize: function alfresco_layout_FixedHeaderFooter___resize() {
-         if (this.recalculateAutoHeightOnResize === true)
+      onResize: function alfresco_layout_FixedHeaderFooter__onResize() {
```
MartinDoyleUK: This naming suggests actions to occur after resize has happened, but it looks like it's actually doing the resize? Can it be called `resize()`, or even `doResize()` if there's a naming clash?

draperd: But that's exactly what is happening... this is called when something has resized, e.g. the window has resized, or an enclosing widget has resized - this in turn triggers the FixedHeaderFooter to resize.

PR #464, opened 2015-08-14
AKU-509: AvatarThumbnail encoding

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-509 to ensure that user names containing the % symbol are encoded before being used in the avatar image URL

### DIFF
--- a/aikau/src/test/resources/alfresco/renderers/AvatarThumbnailTest.js
+++ b/aikau/src/test/resources/alfresco/renderers/AvatarThumbnailTest.js
@@ -50,9 +50,8 @@ define(["intern!object",
       "Publishes topic when clicked": function() {
          return browser.findByCssSelector("#GUEST_THUMBNAIL .inner")
             .click()
-            .getLastPublish("ALF_DISPLAY_NOTIFICATION", true)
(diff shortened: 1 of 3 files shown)


aikau/src/test/resources/alfresco/renderers/AvatarThumbnailTest.js
```diff
@@ -21,45 +21,44 @@
...
-               .then(function(payload) {
-                  assert.propertyVal(payload, "message", "You clicked on the guest thumbnail", "Did not publish correct payload");
-               });
-         },
+      "Publishes topic when clicked": function() {
+         return browser.findByCssSelector("#GUEST_THUMBNAIL .inner")
+            .click()
+            .getLastPublish("ALF_DISPLAY_NOTIFICATION", true)
+            .then(function(payload) {
+               assert.isNotNull(payload, "Did not publish correct topic when clicked");
```
MartinDoyleUK: Not necessary. If you want to publish a message on error, include it as a string argument in the getLastPublish(), which will fail if it cannot find the specified publish, not return a null payload.

PR #465, opened 2015-08-14
AKU-522: Add configuration to stop lists from loading immediately

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-522 to provide a configurable way in which to stop lists from publishing requests to load data as soon as the list is created.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfHashList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfHashList.js
@@ -184,6 +184,7 @@ define(["dojo/_base/declare",
             else
             {
                // No action required, wait for either a reload request or data to be published
+               this.alfLog("log", "This list configured to not load data immediately", this);
             }
          }
       },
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/lists/AlfList.js
```diff
@@ -411,10 +422,14 @@ define(["dojo/_base/declare",
...
          }
-         else
+         else if (this.loadDataImmediately)
          {
             this.loadData();
          }
+         else
+         {
+            // No action required, wait for either a reload request or data to be published
+         }
```
davidcognite: I'm not a fan of empty else blocks - that's how code fails silently. Is it worth at least adding a log message here?

aikau/src/main/resources/alfresco/lists/AlfHashList.js
```diff
@@ -177,10 +177,14 @@ define(["dojo/_base/declare",
...
             }
-            else
+            else if (this.loadDataImmediately)
             {
                this.loadData();
             }
+            else
+            {
+               // No action required, wait for either a reload request or data to be published
+            }
```
davidcognite: empty else block = log message?

PR #474, opened 2015-08-18
AKU-494: Update doclib.lib.js to use useHash option on PathTree config

draperd:
Update to doclib.lib.js to pass on useHash from options to PathTree (see https://issues.alfresco.com/jira/browse/AKU-494)

### DIFF
--- a/aikau/src/main/resources/webscript-libs/doclib/doclib.lib.js
+++ b/aikau/src/main/resources/webscript-libs/doclib/doclib.lib.js
@@ -577,7 +577,8 @@ function getDocLibTree(options) {
                   siteId: options.siteId,
                   containerId: options.containerId,
                   rootNode: options.rootNode,
-                  rootLabel: options.rootLabel || "documentlibrary.root.label"
+                  rootLabel: options.rootLabel || "documentlibrary.root.label",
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/webscript-libs/doclib/doclib.lib.js
```diff
@@ -577,7 +577,8 @@ function getDocLibTree(options) {
                   siteId: options.siteId,
                   containerId: options.containerId,
                   rootNode: options.rootNode,
-                  rootLabel: options.rootLabel || "documentlibrary.root.label"
+                  rootLabel: options.rootLabel || "documentlibrary.root.label",
+                  useHash: (options.useHash !== false)
```
MartinDoyleUK: Does the `options` object always have a `useHash` property? Otherwise, it could be `undefined` and it would pass through `true`? Or is that the intention? How about the string `"false"`?

draperd: This is the same as all the other places that options.useHash is used (e.g in the breadcrumb trail and the document list)... we want the default to be true, so yes... useHash could be undefined in which case we are happy that useHash is set to true.

PR #481, opened 2015-08-19
AKU-515: Correct invalid falsy logic for global/parent publish

draperd:
This addresses https://issues.alfresco.com/jira/browse/AKU-515 to ensure that pubSubScope is respected in the alfresco/renderers/Thumbnail widget. The unit test has been updated to verify scoping.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/Thumbnail.js
+++ b/aikau/src/main/resources/alfresco/renderers/Thumbnail.js
@@ -465,8 +465,8 @@ define(["dojo/_base/declare",
          }
          else
          {
-            var publishGlobal = this.publishGlobal !== false;
-            var publishToParent = this.publishToParent !== false;
+            var publishGlobal = this.publishGlobal || false;
+            var publishToParent = this.publishToParent || false;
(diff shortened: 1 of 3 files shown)


aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -465,8 +465,8 @@ define(["dojo/_base/declare",
          }
          else
          {
-            var publishGlobal = this.publishGlobal !== false;
-            var publishToParent = this.publishToParent !== false;
+            var publishGlobal = this.publishGlobal || false;
+            var publishToParent = this.publishToParent || false;
             this.alfPublish(this.publishTopic, this.publishPayload, publishGlobal, publishToParent);
```
MartinDoyleUK: Could all be written on one line, if desired, including forcing to boolean (is that really required??) ...

```
this.alfPublish(this.publishTopic, this.publishPayload, !!this.publishGlobal, !!this.publishToParent);
```

draperd: This boils down to personal preference... I find mentally processing multiple exclamation marks problematic as they're easily missed, what we have now is very clearly defined.

PR #484, opened 2015-08-20
AKU-523: Warnings in forms support

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-523 to add support for configurable warnings in forms. The warnings are processed using the same rules engine that was originally in the BaseFormControl (the code has now been abstracted to a RulesEngineMixin module so that it can be shared with the Form widget). I've made updates to alfresco/header/Warning to add support for dynamic visibility over pub/sub and have applied BEM to it. 

The unit tests added ensure that the warnings work for both regular forms and forms in dialogs. The warning can also be styled via LESS variables.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/utilities/RulesEngineMixin.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/utilities/RulesEngineMixin.js
@@ -41,7 +41,7 @@ define(["dojo/_base/declare",
        *
        * @instance
        * @type {object}
-       * @default null
+       * @default
        */
       _rulesEngineData: null,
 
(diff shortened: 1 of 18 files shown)


aikau/src/main/resources/alfresco/forms/controls/utilities/RulesEngineMixin.js
```diff
@@ -0,0 +1,305 @@
...
+
+   return declare([], {
+
+      /**
+       * This holds all the data about rules that need to be processed for the various attributes of the widget. By default this
+       * will handle rules for visibility, requirement and disability.
+       *
+       * @instance
+       * @type {object}
+       * @default null
```
MartinDoyleUK: Don't need to specify nulls (and if it changes later, this default could rust)

MartinDoyleUK:
Long PR ... can't see any problems with it, although it is quite large. Happy to merge when you're ready.

draperd:
Updated the JSDoc defaults, please merge.

PR #485, opened 2015-08-21
AKU-528: Further link updates

draperd:
This PR makes further updates to https://issues.alfresco.com/jira/browse/AKU-528. Primarily the purpose was to add in additional LESS configuration for link hover states to allow themes to be customized so that different decoration and colour can be applied based on whether or not the mouse is hovering over the link or not. 

I've also made further updates around the alfresco/header/Title to improve accessibility by ensuring that title are only links when they have a targetUrl set and defining dedicated LESS variables for title links (because I think they should be special cased).

Default Aikau LESS variable remain consistent with Share 5.0.x (e.g no text decoration or colours for links).

### DIFF
--- a/aikau/src/main/resources/alfresco/header/Title.js
+++ b/aikau/src/main/resources/alfresco/header/Title.js
@@ -25,22 +25,30 @@
  * 
  * @module alfresco/header/Title
  * @extends external:dijit/_WidgetBase
- * @mixes external:dojo/_TemplatedMixin
+ * @mixes external:dijit/_TemplatedMixin
+ * @mixes external:dijit/_OnDijitClickMixin
+ * @mixes module:alfresco/navigation/_HtmlAnchorMixin
+ * @mixes module:alfresco/services/_NavigationServiceTopicMixin
  * @mixes module:alfresco/core/Core
  * @author Dave Draper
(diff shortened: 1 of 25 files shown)


aikau/src/main/resources/alfresco/header/Title.js
```diff
@@ -129,9 +151,53 @@ define(["dojo/_base/declare",
...
+       * Handles click events when a [targetUrl]{@link module:alfresco/header/Title#targetUrl} has
+       * been provided.
+       *
+       * @instance
+       * @param {object} evt The click event
+       * @since 1.0.32
+       */
+      onClick: function alfresco_header_Title__onClick(evt) {
+         var targetUrlLocation = this.targetUrlLocation;
+         if (has("mac") && evt.metaKey)
```
MartinDoyleUK: Would it be more appropriate to use `evt.copyKey` here?

draperd: I don't know - I took the existing code from AlfMenuItemMixin... I didn't add it originally, both should be consistent - there is no compelling reason to change it at the moment (certainly not as part of this issue) but possibly something to consider under a different issue

PR #486, opened 2015-08-21
AKU-531: Added service registry

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-531 to provide a simple way in which to prevent multiple instances of the same service from being registered on the same pubSubScope. It introduces a singleton alfresco/services/ServiceRegistry (that maintains registration of services) and the alfresco/services/BaseService (that all core Aikau services now extend).

This ensures that where we have more than one Surf Component on a page registering Aikau content that duplicate services will be weeded out. This is a better approach than using the previously recommended service-filter.lib.js as it happens transparently. 

A unit test has been created to verify this works.


draperd:
Please review, but hold off on merging until I've done some more testing on Share and run the full regression suite.

aikau/src/main/resources/alfresco/services/BaseService.js
```diff
@@ -0,0 +1,61 @@
...
+      
+      /**
+       * Sets up the subscriptions for the LogoutService
+       * 
+       * @instance
+       * @param {array} args Constructor arguments
+       */
+      constructor: function alfresco_services_BaseService__constructor(args) {
+         declare.safeMixin(this, args);
+         if (ServiceRegistry.getSingleton().register(args.alfServiceName, args.pubSubScope))
```
MartinDoyleUK: Would prefer to read `this.alfServiceName, this.pubSubScope` as it's more consistent with how these are normally referenced, although I realise this is subjective to say the least!

aikau/src/main/resources/alfresco/services/LightboxService.js
```diff
@@ -19,16 +19,16 @@
...
  * @module alfresco/services/LightboxService
- * @extends module:alfresco/core/Core
+ * @extends module:alfresco/services/BaseService
  * @author Kevin Roast
  */
 define(["dojo/_base/declare",
-        "alfresco/core/Core",
+        "alfresco/services/BaseService",
         "dojo/_base/lang",
         ""],
```
MartinDoyleUK: It's still there.... ??

aikau/src/main/resources/alfresco/services/BaseService.js
draperd: Are they actually available on "this" at that point though - I'm not sure that they are, although I can double check.

aikau/src/main/resources/alfresco/services/LightboxService.js
draperd: Bugger, I removed it and then discarded changes and forgot to remove it again... this is why we have these nice code reviews :)

draperd:
I've now verified this works as it should in Share.... just pending a potential update on how we handle the singleton pattern and whether or not we need it as a class

draperd:
It looks like we still need a constructor on each of the services anyway... I'll need to push some more changes next week before we can merge this so there's more time for discussion on the finer details.

draperd:
Unit tests are now all passing - this is ready to merge from a functional point of view, it sounds like we just want some agreement on that logging message now.

davidcognite:
I'm happy to put aside the config issue and get this code merged.

PR #487, opened 2015-08-21
AKU-526: Add support for infinite scrolling in FilmStripView

MartinDoyleUK:
This pull request primarily addresses [AKU-526](https://issues.alfresco.com/jira/browse/AKU-526) and adds support for infinite scrolling in FilmStripView. Additionally, DebugLog has been updated to add an exclude-filter (in addition to the include-filter) and to allow selecting/copying of the payloads and a new mock XHR class has been created which will mock the document library on a large, generic scale. Finally, a few tests were made more robust, which were failing with timing issues on Chrome.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -90,6 +90,23 @@ define([],function() {
        * @type {string}
        * @default
        */
-      UPDATE_PAGE_TITLE: "ALF_UPDATE_PAGE_TITLE"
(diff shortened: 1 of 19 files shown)


draperd:
I can't see how the AlfFilmStripView is able to distinguish whether or not it is being used in infiniteScroll mode? I would have liked to have seen a separate test page verifying both states. We need to be able to support standard pagination (as was the previous behaviour) so that you can jump ahead if there are a large number of pages.

I'd like to see a new test page that uses the view with a list configured to use infinite scrolling and retain the existing test that used regular pagination.

PR #491, opened 2015-08-25
AKU-526: Filmstrip view with infinite scroll corrections

draperd:
This PR makes some updates following the initial PRs for https://issues.alfresco.com/jira/browse/AKU-526. After some preliminary testing in Share we hit a few issues that needed to be resolved because the AlfSearchList was not calling the processLoadedData function so wasn't taking advantage of some of the improvements made in the previous PR. This has now been corrected along with some defensive code updates and some other minor corrections

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/views/layouts/Carousel.js
+++ b/aikau/src/main/resources/alfresco/lists/views/layouts/Carousel.js
@@ -156,9 +156,11 @@ define(["dojo/_base/declare",
        */
       resize: function alfresco_lists_views_layouts_Carousel__resize() {
          this.calculateSizes();
-         domStyle.set(this.itemsNode, "width", this.itemsNodeWidth + "px");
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/lists/views/layouts/Carousel.js
```diff
@@ -372,9 +374,9 @@ define(["dojo/_base/declare",
        * @param {object} payload
        */
       selectItem: function alfresco_lists_views_layouts_Carousel__item(payload) {
-         if (payload.index != null && !isNaN(parseInt(payload.index)))
+         if ((payload.index || payload.index === 0) && !isNaN(parseInt(payload.index, 10)))
```
MartinDoyleUK: Maybe have ...

```
var targetIndex = payload.index && parseInt(payload.index, 10);
if(!isNaN(targetIndex)) {
```

... which means you can also delete the line below as well.

PR #501, opened 2015-08-27
Fix spelling mistake

MartinDoyleUK:


aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/lists/LocalStorageFallback.get.js
```diff
@@ -120,4 +120,4 @@ model.jsonModel = {
          name: "alfresco/logging/DebugLog"
       }
    ]
-};
\ No newline at end of file
+};
```
MartinDoyleUK: Edited this file directly on GitHub and it changed this line on its own!

PR #503, opened 2015-08-27
AKU-478: Horizontal Widget Resizing

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-478 to update the alfresco/layout/HorizontalWidgets to take the visibilityConfig and invisibilityConfig of it's child widgets into consideration when calculating the widths of those widgets. It sets up a subscription for each unique topic used across all the visibility rules where each subscription is bound to the onResize function. This means that each time a child of a HorizontalWidgets changes visibility the widths of all the children will be re-calculated to ensure that all the available horizontal space is consumed without overflow.

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/HorizontalWidgets.js
+++ b/aikau/src/main/resources/alfresco/layout/HorizontalWidgets.js
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2005-2014 Alfresco Software Limited.
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
  *
  * This file is part of Alfresco
  *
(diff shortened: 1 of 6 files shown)


MartinDoyleUK:
This sounds a lot like Flexbox by JavaScript! :)

aikau/src/main/resources/alfresco/layout/HorizontalWidgets.js
```diff
@@ -128,6 +138,70 @@ define(["alfresco/core/ProcessWidgets",
...
+         });
+
+         for (var key in ruleTopicsObject) 
+         {
+            if (ruleTopicsObject.hasOwnProperty(key))
+            {
+               topics.push(key);
+            }
+         }
+         return topics;
```
MartinDoyleUK: This is a very verbose function. It could be written more concisely ...

```javascript
var topicNames = {};
array.forEach(widgets, function(widget) {
  var visibilityRules = lang.getObject("config.visibilityConfig.rules", false, widget) || [],
    invisibilityRules = lang.getObject("config.invisibilityConfig.rules", false, widget) || [];
  array.forEach(visibilityRules.concat(invisibilityRules), function(rule) {
    if (rule.topic) {
      topicNames[rule.topic] = true;
    }
  });
});
return Object.keys(topicNames);
```

... if one wanted to, of course!

PR #507, opened 2015-08-28
AKU-532: Further download as zip updates

draperd:
This PR is some further updates related to https://issues.alfresco.com/jira/browse/AKU-532. In particular this addresses support when "rawData" is used (e.g. using APIs directly on the Repository rather than the Share tier), has a standalone client icon rendering fallback and ensures that selected items are cleared when the list data is cleared (e.g. to support the use in faceted search when applying facets). I've also included some general improvements to the doclib.lib.js file.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfList.js
@@ -734,6 +734,9 @@ define(["dojo/_base/declare",
        * @instance
        */
       clearViews: function alfresco_lists_AlfList__clearViews() {
+         // Publish the clear selected items topic to ensure that any selected items menus
+         // don't retain stale data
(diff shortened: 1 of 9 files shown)


aikau/src/main/resources/alfresco/lists/AlfList.js
```diff
@@ -734,6 +734,9 @@ define(["dojo/_base/declare",
        * @instance
        */
       clearViews: function alfresco_lists_AlfList__clearViews() {
+         // Publish the clear selected items topic to ensure that any selected items menus
+         // don't retain stale data
+         this.alfPublish(topics.CLEAR_SELECTED_ITEMS, {});
```
MartinDoyleUK: Empty object literal not necessary (the function handles it not being present automatically)

PR #511, opened 2015-09-02
Removed obsolete negative margin from Warning component (used in Alfresco Share)

kevinroast:
Since the warning component is now used to display Share Services AMP version mismatch, it is more visible and shows that an old margin setting is still present - not needed and adds an unwanted horizontal scrollbar.

### DIFF
--- a/aikau/src/main/resources/alfresco/header/css/Warning.css
+++ b/aikau/src/main/resources/alfresco/header/css/Warning.css
@@ -7,7 +7,6 @@
    }
 
    .alfresco-header-Warning__warnings {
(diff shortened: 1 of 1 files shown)


draperd:
For context, I recall this was added because of the #doc3 CSS selector in Share that sets a 10px margin. This meant that the warning on the Share dashboard (where this widget was originally written to be shown) would not reach the edges of the page. What we probably need is a more specific selector just to address this scenario as well. Full page Aikau widgets don't have this margin so this would explain the horizontal scrollbar.

PR #517, opened 2015-09-04
AKU-565: Use 20 X 20 px Delete Icon

ahmedowian:
Please review this pull request for AKU-565: Use 20 X 20 px Delete Icon.


ahmedowian:
I made this by mistake.

PR #518, opened 2015-09-04
AKU-565: Use 20 X 20 px Delete Icon

ahmedowian:
- Added delete-default-20.png which is 20 X 20 px
   - Updated alf-delete-icon to use new image


draperd:
Before this gets merged by anyone I want to fully review all the things that might be impacted by this change. As the previous icon is almost certainly used in lots of places. What we actually might want is an easier way to customize icons.

Not least this is going to be used in the 5.0.x versions of Share for the faceted search actions and we cannot allow that to be changed without good reason.

draperd:
This can't be merged as is. Closing. The "bug" has been converted to a feature.

PR #520, opened 2015-09-07
AKU-538: FilmStripView preview height

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-538 to ensure that the previews in the AlfFilmStripView aren't set an inappropriate height when the browser loads halfway down the page. This has been resolved by adding a new alfresco/layout/HeightMixin module that has been mixed into the alfresco/views/layouts/Carousel (the offending widget) as well as the alfresco/preview/AlfDocumentPreviewPlugin (from which the mixin code was taken).

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/HeightMixin.js
+++ b/aikau/src/main/resources/alfresco/layout/HeightMixin.js
@@ -46,21 +46,21 @@ define(["dojo/_base/declare",
    return declare([], {
       
       /**
-       * This property allows the height of the sidebar to accommodate a "sticky" footer. The height is otherwise calculated as
-       * the height of the view port minus the top position of the DOM element having its height set. By setting this property 
-       * it is possible to also deduct the height of a sticky footer. 
+       * This property allows for the height being set to be compensated for a number of different factors
+       * such as sticky footers or configured padding or margins that might increase the height of the widget
+       * DOM element.
        * 
        * @instance
        * @type {number} 
        * @default
        */
-      footerHeight: 0,
+      heightAdjustment: 0,
       
       /**
        * <p>This should be configured to indicate how the height of the widget should be calculated.
        *   <ul>
        *     <li>"AUTO" (the default) indicates that the height will be calculated to be the available space from the widgets position
-       * to the bottom of the screen (minus any [footerHeight]{@link module:alfresco/layout/HeightMixin#footerHeight}).<li>
+       * to the bottom of the screen (minus any [heightAdjustment]{@link module:alfresco/layout/HeightMixin#heightAdjustment}).<li>
        *     <li>"DIALOG" indicates that the height should be taken from the available height of the dialog in which the widget
        * is displayed.</li>
        *     <li>Any negative number indicates that the "AUTO" height minus the supplied value will be used.</li>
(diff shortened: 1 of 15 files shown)


aikau/src/main/resources/alfresco/layout/HeightMixin.js
```diff
@@ -0,0 +1,129 @@
...
+               // We're using JQuery here to get the offset as it has proved more reliable than either the Dojo margin box
+               // or native browser offsetTop options...
+               var offset = $(domNode).offset().top;
+               var availableHeight = winBox.h - offset - this.footerHeight;
+               calculatedHeight =  availableHeight;
+            }
+            else if (heightMode < 0)
+            {
+               // If the height mode is a number less than zero, then deduct that height from the available space.
+               calculatedHeight = winBox.h + heightMode; // NOTE: Not a mistake, remember adding a negative number substracts! :)
```
davidcognite: This method doesn't follow the JSDoc explanation of what happens if you specify a negative number in heightMode. JSDoc says it'll be the same as AUTO, but with the heightMode subtracted, however the calculation for AUTO is different than just winBox.h

aikau/src/main/resources/alfresco/lists/views/layouts/Carousel.js
```diff
@@ -223,13 +224,7 @@ define(["dojo/_base/declare",
...
             else
             {
-               // Calculate a suitable height...
-               // We're going to set the height to be as big as the viewing port allows
-               // from the starting vertical position of the node
-               var position = domGeom.position(this.domNode);
-               var viewPort = win.getBox();
-               this.height = viewPort.h - position.y;
-               this.height += "px";
+               this.height = this.calculateHeight(this.domNode) + "px";
```
davidcognite: Appending px is going to be a common use case - worth including within the calculateHeight method? (controlled via a boolean param & only added if calculatedHeight isn't undefined?)

aikau/src/test/resources/alfresco/documentlibrary/views/FilmStripViewTest.js
```diff
@@ -629,8 +629,25 @@ define([
...
+            .clearLog()
+
+            .findByCssSelector("#PAGED_FILMSTRIP_VIEW_PREVIEWS li:first-child .alfresco-renderers-Thumbnail .inner img")
+               .click()
+            .end()
+
+            .getLastPublish("PAGED_ALF_RETRIEVE_DOCUMENTS_REQUEST_SUCCESS", "Did not load data");
          },
 
          "Post Coverage Results": function() {
```
davidcognite: Test doesn't really provide much coverage on the HeightMixin, does it?

draperd: This test wasn't possible before the mixin code was updated and the existing preview tests also cover it as well, however you are right that there isn't a dedicate test for the mixin itself.

aikau/src/main/resources/alfresco/layout/HeightMixin.js
draperd: Yep, good spot ... lazy copy paste on my part.

aikau/src/main/resources/alfresco/lists/views/layouts/Carousel.js
draperd: I'm not sure it is going to be that common a use case. It would probably be recommended to just call "setHeight", however I added the "calculateHeight" function to address the way this widget was originally written as I didn't want to change it too much initially.

draperd:
Don't merge this yet - I've found more things that need to be corrected.

draperd:
I've closed this off and recreated as https://github.com/Alfresco/Aikau/pull/521

PR #521, opened 2015-09-08
AKU-538: Created HeightMixin module

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-538 to primarily ensure that the preview carousel in the FilmStripView is the correct height but also provides a new mixin that can be used for setting height of widgets. Tests have been updated and tweaks have been made to the way that dialogs and previews are displayed in order to get the optimum sizes.


davidcognite:
Looks good to me. I think the jQuery selectors could probably be optimised a little, but that's just nitpicking, so I'm going to merge anyway.

PR #531, opened 2015-09-11
AKU-551: Retain item selection when switching views

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-551 to ensure that when switching views any item selection on one view is retained on the next.

### DIFF
--- a/aikau/src/main/resources/alfresco/menus/AlfMenuBarSelectItems.js
+++ b/aikau/src/main/resources/alfresco/menus/AlfMenuBarSelectItems.js
@@ -201,6 +201,10 @@ define(["dojo/_base/declare",
                this._itemsSelected = this._UNKNOWN;
             }
          }
+         else
+         {
+            this.alfLog("warn", "Invalid item selection requested", payload, this);
+         }
       },
       
       /**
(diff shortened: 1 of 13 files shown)


aikau/src/main/resources/alfresco/menus/AlfMenuBarSelectItems.js
```diff
@@ -176,19 +176,22 @@ define(["dojo/_base/declare",
        * @param {object} payload
        */
       determineSelection: function alfresco_menus_AlfMenuBarSelectItems__determineSelection(payload) {
-         if (payload.availableItemCount != null && payload.selectedItemCount != null) 
+         var available = parseInt(payload.availableItemCount, 10);
```
tunaaksoy: if payload.availableItemCount or payload.selectedItemCount will be null or undefined NaN will be returned.

draperd: ...but then they'll just fail the next if block? Which I think will be fine?

tunaaksoy: Ok, that's true. Would it make sense to log this then in an else statement?

draperd: Yes, I can make that change

PR #537, opened 2015-09-15
AKU-559: Support async widget processing callbacks

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-559 and addresses the issue of async order when calling processWidgets that can occur when JS dependencies are missing. In particular this occurs where Surf has not been able to include locale specific dependencies (so this problem only occurs when the browser is using the non-default locale). It adds support for mapping the final result of a call to processWidgets with a specific ID of that call.


MartinDoyleUK:
All looks fine to me

PR #539, opened 2015-09-16
Extracted top-level CSS selector as LESS variable

kevinroast:
This change allows Aikau app developers the freedom to change the top-level selector from "alfresco-share" to something else and still make use of the standard Aikau widget CSS.

The top-level selector is now defined as a LESS default in src/main/resources/alfresco/css/less/defaults.less:
```css
// Root selector
@alfresco: alfresco-share;
```

In a custom theme an example override would be:
```xml
<?xml version='1.0' encoding='UTF-8'?>
<theme>
   <id>sfs-theme</id>
   <title>SFS Theme</title>
   <css-tokens>
      <less-variables>
         @alfresco: somevalue;
      </less-variables>
   </css-tokens>
</theme>
```

NOTE! This has a dependency on fixing the compression of CSS files in Surf - this is currently achieved with the following *mandatory* application context bean override to remove *.css from compression step - example for Alfresco Share:
```xml
   <bean id="dependency.aggregator" parent="dependency.aggregator.abstract" class="org.springframework.extensions.surf.DependencyAggregator">
      <property name="compressionExclusions">
          <list>
             <value>*.css</value>
             <value>*-min.js</value>
             <value>*/lib/code-mirror/*.js</value>
          </list>
       </property>
   </bean>
```
The `*.css` is the important line. This is *required* or LESS errors will occur with this change!

### DIFF
--- a/aikau-archetype/src/main/resources/archetype-resources/src/main/webapp/WEB-INF/config/web-application-config.xml
+++ b/aikau-archetype/src/main/resources/archetype-resources/src/main/webapp/WEB-INF/config/web-application-config.xml
@@ -137,4 +137,15 @@
          </list>
       </property>
    </bean>
+   
+   <!-- Override DependencyAggregator to set compression exclusions -->
+   <bean id="dependency.aggregator" parent="dependency.aggregator.abstract" class="org.springframework.extensions.surf.DependencyAggregator">
+      <property name="compressionExclusions">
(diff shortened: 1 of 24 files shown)


draperd:
I've checked out the branch and built it locally to verify that the unit test application works as expected (which it does) so I'll merge

PR #541, opened 2015-09-16
AKU-574: Sidebar contents hidden on collapse

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-574. I've gone through the implementation to use LESS with a BEM approach. But the fix is somewhat hidden in that all the children of the sidebar element (apart from the resize bar) have display:none applied when the sidebar is closed. This means that they cannot prevent the open/close toggle from being used. But as an added measure I've increased the z-index as well (making sure that it is less than the dialog z-index)

### DIFF
--- a/aikau/src/main/resources/alfresco/layout/AlfSideBarContainer.js
+++ b/aikau/src/main/resources/alfresco/layout/AlfSideBarContainer.js
@@ -379,7 +379,7 @@ define(["dojo/_base/declare",
             // not. We're using the classes applied to the drag handle node to determine whether or not
             // to show or hide the sidebar.
             this.alfPublish("ALF_DOCLIST_SHOW_SIDEBAR", {
-               selected: domClass.contains(this.resizeHandlerButtonNode, "pop-out")
+               selected: domClass.contains(this.sidebarNode, "alfresco-layout-AlfSideBarContainer__sidebar--closed")
             });
          }
       },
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/layout/AlfSideBarContainer.js
```diff
@@ -407,42 +407,23 @@ define(["dojo/_base/declare",
...
             this.lastSidebarWidth = width;
             this.resizeHandler();
-            if (this.resizeHandlerButtonNode)
-            {
-               domClass.remove(this.resizeHandlerButtonNode, "pop-out");
-               domClass.add(this.resizeHandlerButtonNode, "pop-in");
-            }
+
+            domClass.add(this.sidebarNode, "alfresco-layout-AlfSideBarContainer__sidebar--open");
+            domClass.remove(this.sidebarNode, "alfresco-layout-AlfSideBarContainer__sidebar--closed");
```
MartinDoyleUK: This should be a single class that's added/removed as necessary whenever the states are mutually exclusive. There's no real convention that I'm aware of for how to pick which one, but it makes sense that the defined state is the non-default one, so presumable that'd mean using `--closed`.

When using a boolean to determine the state-change, I find it tidiest to do this ...

```javascript
domClass[bool ? "add" : "remove"](node, className);
```

... but I accept that might be a bit too concise for some!

aikau/src/test/resources/alfresco/layout/AlfSideBarContainerTest.js
```diff
@@ -61,7 +61,7 @@ define(["intern!object",
       },
 
       "Test Sidebar Placement": function () {
-         return browser.findByCssSelector(".alfresco-layout-AlfSideBarContainer .sidebar #SIDEBAR_LOGO")
+         return browser.findByCssSelector(".alfresco-layout-AlfSideBarContainer .alfresco-layout-AlfSideBarContainer__sidebar #SIDEBAR_LOGO")
```
MartinDoyleUK: Because of BEM, there's no need to nest rules, so we should be able to get rid of the parent `.alfresco-layout-AlfSideBarContainer` in all of the examples in this file.

PR #545, opened 2015-09-16
AKU-584: Ensure correct scope on SearchService XHR

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-584 to ensure that the SearchService uses the correct scope when making XHR requests. The unit tests have been updated to prevent future regressions.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/SearchService.js
+++ b/aikau/src/main/resources/alfresco/services/SearchService.js
@@ -32,9 +32,10 @@ define(["dojo/_base/declare",
         "service/constants/Default",
         "alfresco/core/PathUtils",
         "alfresco/core/NodeUtils",
+        "alfresco/core/topics",
         "dojo/_base/lang",
         "dojo/json"],
-        function(declare, BaseService, CoreXhr, AlfConstants, PathUtils, NodeUtils, lang, dojoJson) {
+        function(declare, BaseService, CoreXhr, AlfConstants, PathUtils, NodeUtils, topics, lang, dojoJson) {
 
    return declare([BaseService, CoreXhr, PathUtils], {
 
@@ -167,6 +168,7 @@ define(["dojo/_base/declare",
          if (!payload || !payload.term)
          {
             this.alfLog("warn", "A request was made to perform a search but no 'term' attribute was provided", payload, this);
+            this.alfPublish(payload.alfResponseScope + payload.alfResponseTopic + "_FAILURE");
          }
          else
          {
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/services/SearchService.js
```diff
@@ -167,6 +168,7 @@ define(["dojo/_base/declare",
          if (!payload || !payload.term)
          {
             this.alfLog("warn", "A request was made to perform a search but no 'term' attribute was provided", payload, this);
+            this.alfPublish(payload.alfResponseScope + payload.alfResponseTopic + "_FAILURE");
```
MartinDoyleUK: We should avoid manually adding a scope to a publish, as the resultant payload will pick up the default scope and use that (even if it's correct in this instance). A specific scope should be provided in the call the publish, to ensure consistency of behaviour.

```javascript
this.alfPublish(payload.alfResponseTopic + "_FAILURE", {}, false, false, payload.alfResponseScope);
```

If this appears too unwieldy, the solution would probably be to later update the `alfPublish()` signature.

aikau/src/test/resources/alfresco/services/SearchServiceTest.js
```diff
@@ -84,6 +84,87 @@ define(["intern!object",
...
+         .end()
+         .findById("FCTSRCH_SET_ALL_SITES_SCOPE_text")
+            .click()
+         .end()
+         .getLastPublish("ALF_SEARCH_REQUEST")
+            .then(function(payload) {
+               assert.propertyVal(payload, "repo", false, "Repository scope used");
+               assert.propertyVal(payload, "site", "", "Site incorrect");
+            })
+         .end()
```
MartinDoyleUK: `end()` not necessary here as getLastPublish (and similar methods) do not change the CSS scope

aikau/src/test/resources/testApp/js/aikau/testing/mockservices/SearchMockXhr.js
```diff
@@ -39,10 +39,11 @@ define(["dojo/_base/declare",
          {
             var templateToUse = xhrSearchResponse;
             this.server.respondWith("GET",
-                                    /\/aikau\/proxy\/alfresco\/slingshot\/search/,
+                                    /(.*)/,
                                     [200,
                                      {"Content-Type":"application/json;charset=UTF-8"},
                                      templateToUse]);
+            this.alfPublish("ALF_MOCK_XHR_SERVICE_READY", {});
```
MartinDoyleUK: Object literal not necessary, as method defaults the optional payload to the same

PR #554, opened 2015-09-21
Increase timeouts to improve test consistency

draperd:
I've found that by increasing the time outs that the unit tests pass more reliable when run in bulk.


MartinDoyleUK:
This needs to be a temporary fix. Arbitrary timeouts and increasing them should never be a fix. We need to find more reliable methods of ensure events have happened before we try to test them.

PR #563, opened 2015-09-23
AKU-588: Ensure resize events published on dynamic width resizing

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-588. It ensures that resize publications are made from the alfresco/layout/HorizontalWidgets module when widths are dynamically updated. This will be merged to develop and then cherry picked for a 1.0.36.1 hotfix release.

### DIFF
--- a/aikau/src/test/resources/alfresco/layout/DynamicHorizontalLayoutTest.js
+++ b/aikau/src/test/resources/alfresco/layout/DynamicHorizontalLayoutTest.js
@@ -285,12 +285,11 @@ define(["intern!object",
          },
 
          "Check that when a widget is revealed that a resize event is published": function() {
-            return browser.findByCssSelector("body").end()
+            return browser.findById("HIDE_WIDGET_2_BUTTON_label")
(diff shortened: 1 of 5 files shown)


aikau/src/test/resources/alfresco/layout/DynamicHorizontalLayoutTest.js
```diff
@@ -283,6 +283,15 @@ define(["intern!object",
                   assert.closeTo(size.width, expectedWidth, 10, "Widget 5 was not the correct size");
                });
          },
+
+         "Check that when a widget is revealed that a resize event is published": function() {
+            return browser.findByCssSelector("body").end()
+               .clearLog()
+               .findById("HIDE_WIDGET_2_BUTTON_label")
```
MartinDoyleUK: If this findById was first then you could call clearLog immediately after (and prior to the click) without affecting the CSS scope, and the find/end of the body would be redundant.

PR #567, opened 2015-09-24
AKU-597: Ensure facets from URL hash are applied to search

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-597. Because of changes to the way in which the AlfSearchList is now initialised the facetFilters object is not getting initialised from the hash correctly.

### DIFF
--- a/aikau/src/test/resources/alfresco/documentlibrary/SearchListTest.js
+++ b/aikau/src/test/resources/alfresco/documentlibrary/SearchListTest.js
@@ -27,305 +27,305 @@ define(["intern!object",
         "alfresco/TestCommon"], 
         function (registerSuite, expect, assert, require, TestCommon) {
 
-   // registerSuite(function(){
-   //    var browser;
+   registerSuite(function(){
+      var browser;
 
-   //    return {
-   //       name: "SearchList Tests",
+      return {
+         name: "SearchList Tests",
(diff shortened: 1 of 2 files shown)


aikau/src/test/resources/alfresco/documentlibrary/SearchListTest.js
```diff
@@ -27,303 +27,332 @@ define(["intern!object",
...
+   //          .getLastPublish("ALF_SEARCH_RESULTS_COUNT")
+   //          .then(function(payload) {
+   //             assert.propertyVal(payload, "count", 3, "Search resultset size not published");
+   //          })
+   //          // Test that Number of search results is as expected.
+   //          .findAllByCssSelector(".alfresco-search-AlfSearchResult")
+   //             .then(function(elements) {
+   //                assert.lengthOf(elements, 3, "Number of results expected is 3, actual results displayed is: " + elements.length);
+   //             });
+   //       },
```
MartinDoyleUK: Do we need to commit such a massive amount of commenting-out to version control? Can't we just delete it?

draperd: I don't want to delete it... I forgot to uncomment it :)

PR #570, opened 2015-09-25
AKU-586: alfresco/documentlibrary/AlfTagFilters does not work in a scoped environment

MartinDoyleUK:
This addresses [AKU-586](https://issues.alfresco.com/jira/browse/AKU-586), and deals with the issues with using alfresco/documentlibrary/AlfTagFilters in a scoped environment. It has been tested in Share locally, and new tests have been created and a full test run completed succcesfully. CoreXhr has also been updated to support scoped responses. Additionally, various topics constants have been updated, some small improvements to TestCommon helper methods have been made and the test reporter has had its environment detection/reporting improved.

### DIFF
--- a/aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/documentlibrary/TagFilters.get.html.ftl
+++ b/aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/documentlibrary/TagFilters.get.html.ftl
@@ -1,0 +1,1 @@
+<@processJsonModel group="share"/>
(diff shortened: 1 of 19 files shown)


aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/documentlibrary/TagFilters.get.html.ftl
```diff
@@ -0,0 +1 @@
+<@processJsonModel group="share"/>
```
draperd: Just an FYI... we don't need the group attribute, no need to change - I'm sure there are still lots scattered around in the templates.

PR #578, opened 2015-09-28
AKU-603: Defensively code against duplicate IDs

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-603 to defensively code against duplicate IDs causing failure to process models. This problem was found when creating dialogs (with different IDs) that render form controls that share IDs. Because the dialogIds are different the first dialog is not destroyed so the form controls attempt to use an already registered id. This was particularly problematic with the TabbedControl layout. This has been fixed and the unit tests updated to prevent regressions.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/TabbedControls.js
+++ b/aikau/src/main/resources/alfresco/forms/TabbedControls.js
@@ -83,9 +83,9 @@ define(["alfresco/layout/AlfTabContainer",
         "dojo/_base/declare",
         "dojo/_base/lang",
         "dojo/_base/array"], 
-        function(HorizontalWidgets, LayoutMixin, declare, lang, array) {
+        function(AlfTabContainer, LayoutMixin, declare, lang, array) {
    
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/forms/TabbedControls.js
```diff
@@ -83,9 +83,9 @@ define(["alfresco/layout/AlfTabContainer",
         "dojo/_base/declare",
         "dojo/_base/lang",
         "dojo/_base/array"], 
-        function(HorizontalWidgets, LayoutMixin, declare, lang, array) {
+        function(AlfTabContainer, LayoutMixin, declare, lang, array) {
```
davidcognite: Potential breaking change for anyone who has extended TabbedControls, but has no impact on the actual fix, as far as I can tell, so might be worth leaving out of a hotfix?

draperd: It's not changing the import, just the variable name... anyone extending would have to declare their own variable names so it shouldn't impact them

PR #597, opened 2015-10-08
AKU-548: Cloud Sync Service

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-548 to implement a new service for supporting Cloud Sync operations. In order to complete this task it was necessary to make various changes to core forms behaviour as well as create a new Tree form control.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/CoreXhr.js
+++ b/aikau/src/main/resources/alfresco/core/CoreXhr.js
@@ -313,13 +313,13 @@ define(["dojo/_base/declare",
             this.alfPublish(requestConfig.alfTopic + "_SUCCESS", {
                requestConfig: requestConfig,
                response: response
-            }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
+            }, false, false, requestConfig.alfResponseScope);
          }
          else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
             this.alfPublish(requestConfig.data.alfResponseTopic, {
                requestConfig: requestConfig,
                response: response
-            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
+            }, false, false, requestConfig.data.alfResponseScope);
          }
          else
          {
@@ -341,13 +341,13 @@ define(["dojo/_base/declare",
             this.alfPublish(requestConfig.alfTopic + "_FAILURE", {
                requestConfig: requestConfig,
                response: response
-            }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
+            }, false, false, requestConfig.alfResponseScope);
          }
          else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
             this.alfPublish(requestConfig.data.alfResponseTopic, {
                requestConfig: requestConfig,
                response: response
-            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
+            }, false, false, requestConfig.data.alfResponseScope);
          }
          if (typeof this.displayMessage === "function" && response.response.text)
          {
@@ -390,13 +390,13 @@ define(["dojo/_base/declare",
             this.alfPublish(requestConfig.alfTopic + "_PROGRESS", {
                requestConfig: requestConfig,
                response: response
-            }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
+            }, false, false, requestConfig.alfResponseScope);
          }
          else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
             this.alfPublish(requestConfig.data.alfResponseTopic, {
                requestConfig: requestConfig,
                response: response
-            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
+            }, false, false, requestConfig.data.alfResponseScope);
          }
       },
 
--- a/aikau/src/main/resources/alfresco/core/CoreXhr.js
+++ b/aikau/src/main/resources/alfresco/core/CoreXhr.js
@@ -315,6 +315,12 @@ define(["dojo/_base/declare",
                response: response
             }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
          }
+         else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
+            this.alfPublish(requestConfig.data.alfResponseTopic, {
+               requestConfig: requestConfig,
+               response: response
+            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
+         }
          else
          {
             this.alfLog("warn", "[DEFAULT CALLBACK] Default success callback has been called but no requestConfig.alfTopic has been set.");
@@ -336,6 +342,12 @@ define(["dojo/_base/declare",
                requestConfig: requestConfig,
                response: response
             }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
+         }
+         else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
+            this.alfPublish(requestConfig.data.alfResponseTopic, {
+               requestConfig: requestConfig,
+               response: response
+            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
          }
          if (typeof this.displayMessage === "function" && response.response.text)
          {
@@ -379,6 +391,12 @@ define(["dojo/_base/declare",
                requestConfig: requestConfig,
                response: response
             }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
+         }
+         else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
+            this.alfPublish(requestConfig.data.alfResponseTopic, {
+               requestConfig: requestConfig,
+               response: response
+            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
          }
       },
 
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -99,6 +99,34 @@ define([],function() {
       CLEAR_SELECTED_ITEMS: "ALF_CLEAR_SELECTED_ITEMS",
 
       /**
+       * This can be published to request authentication to the Cloud for the purposes of sync'ing
+       * Alfresco Repository content with the Alfresc Cloud.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       *
+       * @event
+       * @property {string} username The username to authenticate (this should be the e-mail address associated with the Cloud)
+       * @property {string} password The password to authenticate with
+       */
+      CLOUD_AUTHENTICATION_REQUEST: "ALF_CLOUD_AUTHENTICATION_REQUEST",
+
+      /**
+       * This can be published to indicate that authentication to the Cloud for the purposes of sync'ing
+       * Alfresco Repository content with the Alfresc Cloud has completed successfully.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       *
+       * @event
+       */
+      CLOUD_AUTHENTICATION_SUCCESS: "ALF_CLOUD_AUTHENTICATION_SUCCESS",
+
+      /**
        * This topic is published to launch the copying or moving of a node (or nodes) to another location.
        *
        * @instance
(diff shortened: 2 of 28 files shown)


aikau/src/main/resources/alfresco/core/CoreXhr.js
```diff
@@ -315,6 +315,12 @@ define(["dojo/_base/declare",
                response: response
             }, false, false, requestConfig.alfResponseScope || this.pubSubScope);
          }
+         else if (requestConfig.data && requestConfig.data.alfResponseTopic) {
+            this.alfPublish(requestConfig.data.alfResponseTopic, {
+               requestConfig: requestConfig,
+               response: response
+            }, false, false, requestConfig.data.alfResponseScope || this.pubSubScope);
```
MartinDoyleUK: The `|| this.pubSubScope` is implicit inside `alfPublish` if an `undefined` publish scope is provided, i.e. this is redundant and could rust.

Additionally, if there was an explicit `responseScope: ""` in the payload, then this will actually override the intended global response scope and use local scope instead, and so would cause incorrect behaviour.

aikau/src/main/resources/alfresco/core/CoreXhr.js
```diff
@@ -337,6 +343,12 @@ define(["dojo/_base/declare",
...
```
MartinDoyleUK: See above

aikau/src/main/resources/alfresco/core/CoreXhr.js
```diff
@@ -380,6 +392,12 @@ define(["dojo/_base/declare",
...
```
MartinDoyleUK: See above

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -340,6 +403,21 @@ define([],function() {
...
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       * 
+       * @event
+       * @property {string} publishTopic The topic publish the payload once augmented with the form value.
+       */
+      GET_FORM_VALUE_DEPENDANT_OPTIONS: "ALF_GET_FORM_VALUE_DEPENDANT_OPTIONS",
```
MartinDoyleUK: Spelling is dependent

aikau/src/main/resources/alfresco/forms/controls/Tree.js
```diff
@@ -0,0 +1,179 @@
...
+      /**
+       * Sets up the subscriptions to handling tree node selection and then calls 
+       * [createTree]{@link module:alfresco/forms/controls/Tree#createTree} to create the tree to 
+       * be used as the form control.
+       * 
+       * @instance
+       * @param {object} config The configuration to use when instantiating the form control
+       */
+      createFormControl: function alfresco_forms_controls_Tree__createFormControl(/*jshint unused:false*/ config) {
+         this._treeNodeSelectedTopic = this.generateUuid();
```
MartinDoyleUK: We should avoid creating member variables on the fly. They should exist as `null` values and be properly documented.

aikau/src/main/resources/alfresco/forms/controls/Tree.js
```diff
@@ -0,0 +1,179 @@
...
+      },
+
+      /**
+       * This creates a new [tree]{@link module:alfresco/navigation/Tree}.
+       * 
+       * @instance
+       * @return {object} A new tree instance
+       */
+      createTree: function alfresco_forms_controls_Tree__createTree() {
+         return new Tree({
```
MartinDoyleUK: Not using `this.createWidget` intentionally? Would negate the need to manually specify the `pubSubScope` ...

aikau/src/main/resources/alfresco/forms/controls/Tree.js
```diff
@@ -0,0 +1,179 @@
...
+      
+      /**
+       * Currently it is not possible to set a tree node value.
+       * 
+       * @instance
+       * @param {object} value The value to set.
+       */
+      setValue: function alfresco_forms_controls_Tree__setValue(value) {
+         // jshint unused:false
+         // Not possible to set a value currently
```
MartinDoyleUK: Feel we should at least log this when someone tries to do it, if not throwing an error?

aikau/src/main/resources/alfresco/navigation/Tree.js
```diff
@@ -153,6 +153,38 @@ define(["dojo/_base/declare",
...
       
       /**
+       * This is a topic that can be published to request child data.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       */
+      childRequestPublishTopic: null,
```
MartinDoyleUK: This should have an @event tag

aikau/src/main/resources/alfresco/navigation/Tree.js
```diff
@@ -211,6 +243,10 @@ define(["dojo/_base/declare",
 
          // Create a new tree store using the the siteId as part of the URL
          this.treeStore = new TreeStore({
```
MartinDoyleUK: `this.createWidget`?

aikau/src/main/resources/alfresco/navigation/TreeStore.js
```diff
@@ -47,6 +47,49 @@ define(["dojo/_base/declare",
...
 
       /**
+       * This is a topic that can be published to request child data. This an alternative to the 
+       * [target]{@link module:alfresco/navigation/TreeStore#target} URL prefix that can also be configured.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       */
```
MartinDoyleUK: @event tag

aikau/src/main/resources/alfresco/navigation/TreeStore.js
```diff
@@ -47,6 +47,49 @@ define(["dojo/_base/declare",
...
+      /**
+       * Indicates whether or not requests to get child data will be published globally using the 
+       * [publishTopic]{@link module:alfresco/navigation/TreeStore#publishTopic} that has been configured.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.39
+       */
+      publishGlobal: true,
```
MartinDoyleUK: Why not use `_PublishPayloadMixin` which has all of these properties already?

draperd: Because a) They weren't there when I started working on the code, and b) they do different things so the actual documentation is more useful to be duplicated and c) It's not using the actual mixin capabilities - there's no point in pulling in a mixin just for some variables

aikau/src/main/resources/alfresco/navigation/TreeStore.js
```diff
@@ -57,21 +100,66 @@ define(["dojo/_base/declare",
...
+       */
+      onChildren: function alfresco_navigation_TreeStore__onChildren(deferred, parent, subscriptionHandles, payload) {
+         this.alfUnsubscribeSaveHandles(subscriptionHandles);
+         array.forEach(payload.response.items, lang.hitch(this, this.updateChild, parent));
+         
+         // If required, filter the items based on their paths...
+         var filteredResponse = payload.response.items;
+         if (this.filterPaths !== null)
+         {
+            filteredResponse = array.filter(payload.response.items, lang.hitch(this, this.filterChildren));
```
MartinDoyleUK: No point changing this, but could have been written `array.filter(...items, this.filterChildren, this)` without needing to invoke the extra `lang.hitch` function.

aikau/src/main/resources/alfresco/services/CloudSyncService.js
```diff
@@ -0,0 +1,568 @@
...
+      
+      /**
+       * Sets up the subscriptions for the CommentService
+       * 
+       * @instance
+       * @listens module:alfresco/core/topics#CLOUD_AUTHENTICATION_REQUEST
+       * @listens module:alfresco/core/topics#GET_CLOUD_TENANTS
+       * @listens module:alfresco/core/topics#GET_CLOUD_SITES
+       * @listens module:alfresco/core/topics#INIT_CLOUD_SYNC
+       * @listens module:alfresco/core/topics#SYNC_TO_CLOUD
```
MartinDoyleUK: Missing @listens for `GET_CLOUD_PATH`

aikau/src/main/resources/alfresco/services/CloudSyncService.js
```diff
@@ -0,0 +1,568 @@
...
+         // Put the home tenant as the first option...
+         var homeTenant = lang.getObject("homeTenant.name", false, response);
+         options.push({
+            label: homeTenant,
+            value: homeTenant
+         });
+
+         // Add any secondary tenants that have sync-enabled...
+         var secondaryTenants = lang.getObject("secondaryTenants", false, response);
+         array.forEach(secondaryTenants, function(secondaryTenant) {
```
MartinDoyleUK: `dojo/_base/array` actually handles falsy values, but it might be better to maybe have `secondaryTenants || []` just for clarity?

draperd: The API guarantees the array will exist.

aikau/src/main/resources/alfresco/services/CloudSyncService.js
```diff
@@ -0,0 +1,568 @@
...
+                  value: secondaryTenant.name
+               });
+            }
+         });
+
+         // Publish the options back to the originating request...
+         this.alfPublish(originalRequestConfig.data.alfResponseTopic, {
+            requestConfig: originalRequestConfig,
+            response: options
+         }, false, false, originalRequestConfig.data.alfResponseScope || this.pubSubScope);
```
MartinDoyleUK: See comment at top about using the `||` here

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -307,6 +335,41 @@ define([],function() {
...
+       * This topic can be used from a [form control]{@link module:alfresco/form/controls/BaseFormControl}
+       * such as a [Select]{@link module:alfresco/forms/controls/Select} as the optionsConfig.publishTopic
+       * for retrieving the Cloud tenants (networks) that the user has access to.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       */
+      GET_CLOUD_TENANTS: "ALF_GET_CLOUD_TENANTS",
```
MartinDoyleUK: Missing `username` and `remoteTenantId` mandatory properties

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -307,6 +335,41 @@ define([],function() {
...
 
       /**
+       * This topic can be used to request Cloud specific paths to use in an
+       * [Tree form control]{@link module:alfresco/forms/controls/Tree}.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       */
```
MartinDoyleUK: Missing mandatory `remoteSiteId` and `remoteTenantId` properties

aikau/src/main/resources/alfresco/services/CloudSyncService.js
```diff
@@ -0,0 +1,568 @@
...
+         });
+      },
+
+      /**
+       * Publishes a request to show a form [dialog]{@link module:alfresco/dialogs/AlfDialog} containing the 
+       * the [widgetsForSyncDialog]{@link module:alfresco/services/CloudSyncService#widgetsForSyncDialog}
+       * model.
+       * 
+       * @instance
+       * @param {object} payload The payload containing the nodes requested to be sync'd
```
MartinDoyleUK: `payload` suggests that this is the result of a publish, when it's actually just a normal method call?

MartinDoyleUK: Missing @fires for `CREATE_FORM_DIALOG`

draperd: It's always called with a payload... it just happens that this is not the function bound to the subscription.

MartinDoyleUK: Is it better to reserve the `payload` variable name only for functions linked to pub/sub work? To give `payload` a special meaning would make comprehension simpler when reviewing the code (I'd definitely find so. I imagine I'm not alone).

draperd: But the object *is* a payload !!

aikau/src/main/resources/alfresco/navigation/TreeStore.js
MartinDoyleUK: Not sure about (c) but ok :)

PR #598, opened 2015-10-08
Modifications to default theme and widget css to support SFS UX 

spotnape:


aikau/src/main/resources/alfresco/layout/ClassicWindow.js
```diff
@@ -65,6 +66,15 @@ define(["dojo/_base/declare",
       title: "",
 
       /**
+       * Hide the title?
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       */
```
draperd: Missing a @since tag (this e.g. @since 1.0.39 - which will be the next release)

aikau/src/main/resources/alfresco/layout/ClassicWindow.js
```diff
@@ -87,6 +97,10 @@ define(["dojo/_base/declare",
          {
             this.processWidgets(this.widgets, this.contentNode);
          }
+         if (this.hideTitle)
```
draperd: There's no unit test for this new behaviour

aikau/src/main/resources/alfresco/forms/controls/css/TextBox.css
```diff
@@ -18,4 +18,14 @@
...
+}
+
+.alfresco-forms-controls-TextBox.biggerBolder div.control div.dijitValidationTextBox {
+   font-family: @bold-font;
+   font-weight: normal;
+   font-size: @large-font-size;
+}
+
+.alfresco-forms-controls-TextBox.radius div.control div.dijitValidationTextBox {
+   border-radius: @standard-button-border-radius;
```
draperd: This isn't a button. We probably want an additional LESS variable for form control radius that (along with standard-button-border-radius) should be taken from a root seed variable. It should be possible for example to change a variable to remove the radius from a text box without also removing the radius from a button.

spotnape:
Added a test a made other changes accordingly.

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/layout/ClassicWindow.get.js
```diff
@@ -0,0 +1,40 @@
...
+                  name: "alfresco/logo/Logo",
+                  config: {
+                     logoClasses: "alfresco-logo-grey"
+                  }
+               }
+            ]
+         }
+      },
+      {
+         name: "aikauTesting/TestCoverageResults"
```
draperd: This widget is no longer used for test coverage it can be removed, we've been removing them as we touch any existing WebScript test page, so we'd prefer not to add any more (also, unless I mention it this time they'll continue to get added).

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+/**
+ * This is a unit test for ClassicWindow
+ * 
+ * @author Richard Smith
+ */
+define(["intern!object",
+        "intern/chai!assert",
+        "require",
```
draperd: The "require" dependency isn't needed any more, this is another hangover from earlier test pages that we're slowly eliminating.

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
```diff
@@ -0,0 +1,97 @@
...
+         browser.end();
+      },
+
+      "Check the classic windows are present": function () {
+         return browser.findById("WINDOW1")
+            .then(function () {
+               
+            },
+            function () {
+               assert(false, "Classic Window 1 not found");
```
draperd: We typically try to use the appropriate assert function, rather than raw assertions to get better output from the tests, e.g. assert.isFalse

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * @author Richard Smith
+ */
+define(["intern!object",
+        "intern/chai!assert",
+        "require",
+        "alfresco/TestCommon"], 
+        function (registerSuite, assert, require, TestCommon) {
+
+registerSuite(function(){
```
draperd: Could you sort out the indenting here please

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/layout/ClassicWindow.get.html.ftl
```diff
@@ -0,0 +1 @@
+<@processJsonModel group="share"/>
```
draperd: group="share" not needed... again, more legacy copy/paste stuff we'd like to avoid.

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
spotnape: Looking through tests written by the Aikau team I can find a multitude of different indent styles so unfortunately I'm going to need a little more information to solve this one.

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
```diff
@@ -0,0 +1,97 @@
...
+            .end()
+
+            .findById("WINDOW2")
+            .then(function () {
+               
+            },
+            function () {
+               assert(false, "Classic Window 2 not found");
+            });
+      },
```
MartinDoyleUK: I think there's a trick being missed here to make life much easier for us all when writing tests. You've already got a pretty sensible test name so all you need to write is ...

```javascript
"Check that the classic windows are present": function() {
  return browser.findById("WINDOW1")
  .end()
  .findById("WINDOW2");
}
```

... and you can rinse and repeat below as well.

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
draperd: Unfortunately a bulk find replace change in a previous release screwed up some test formatting. Every time I see the incorrect formatting I correct it, so I'd rather not add to the pile of stuff I have to work through and correct.

I would expect everything form HERE .... (END OF INDENTING COMMENT PART 1)

aikau/src/test/resources/alfresco/layout/ClassicWindowTest.js
```diff
@@ -0,0 +1,97 @@
...
+            .getVisibleText()
+            .then(function (text) {
+               assert.strictEqual(text, "Test title");
+            });
+      },
+
+      "Post Coverage Results": function() {
+         TestCommon.alfPostCoverageResults(this, browser);
+      }
+   };
```
draperd: (START OF INDENTING COMMENT PART 2)... to be indented by 3 spaces

draperd:
Thanks for making those changes... it's looking really good now. The only final comment I'd make is that we've added some options that can be used as additionalCssClasses but we're not documenting them at all (e.g. border-less windows, big and bold text boxes) it would be really good if the JSDoc could be updated to list these as options, otherwise how else will developers know that they exist?

spotnape:
Quite happy to do that but not sure of a location or appropriate syntax for such jsdocs. I know that there are similar style features elsewhere too. Any thoughts on an approach to achieving something unified?

@builtInOptions?

draperd:
You're right that at the moment we're not documenting all of them. We're typically updating JSDoc as we touch the modules, at the moment I think the key thing is to capture the information... I'm happy for a simple list to be added in the main widget description, using <ul><li> tags - with the class name in quotes followed by a short description. We can tidy the information up if and when we settle on a standard - we might need to write some more custom JSDoc tags for this in the future.

spotnape:
Jsdocs updated

draperd:
Merged - thanks for making all the changes

PR #600, opened 2015-10-09
AKU-615: Rretain selection on grid resize

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-615. Details of the changes are described in the JIRA ticket, but essentially this moves the selected item state management into a new mixin module and then makes the AlfList the primary manager of item selection state (but allows both the gallery view and selected items popup menu to manage state if configured to do so).

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -673,6 +673,10 @@ define([],function() {
       SELECTED_DOCUMENTS_CHANGED: "ALF_SELECTED_FILES_CHANGED",
 
       /**
+       * This topic should be published by [menu items]{@link module:alfresco/menus/AlfMenuItem} contained within
+       * a [AlfSelectedItemsMenuBarPopup]{@link module:alfresco/documentlibrary/AlfSelectedItemsMenuBarPopup} to
+       * be forwarded onto the [ActionService]{@link module:alfresco/services/ActionService} in order to request
+       * that the configured action be performed on the currently selected items.
        * 
        * @instance
        * @type {string}
@@ -680,7 +684,8 @@ define([],function() {
        * @since 1.0.39
        *
        * @event
-       * 
+       * @param {string} [action] The action to perform
+       * @param {string} [actionTopic] The topic to re-publish the action request on
        */
       SELECTED_DOCUMENTS_ACTION: "ALF_SELECTED_DOCUMENTS_ACTION_REQUEST",
 
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -660,19 +660,6 @@ define([],function() {
       SCROLL_NEAR_BOTTOM: "ALF_SCROLL_NEAR_BOTTOM",
 
       /**
-       * Used to indicate the list of currently selected documents has changed and provides the details of those items.
-       * 
-       * @instance
-       * @type {string} 
-       * @default
-       * @since 1.0.35
-       * 
-       * @event module:alfresco/core/topics~SELECTED_DOCUMENTS_CHANGED
-       * @property {object[]} selectedItems - The items that are selected
-       */
-      SELECTED_DOCUMENTS_CHANGED: "ALF_SELECTED_FILES_CHANGED",
-
-      /**
        * This topic should be published by [menu items]{@link module:alfresco/menus/AlfMenuItem} contained within
        * a [AlfSelectedItemsMenuBarPopup]{@link module:alfresco/documentlibrary/AlfSelectedItemsMenuBarPopup} to
        * be forwarded onto the [ActionService]{@link module:alfresco/services/ActionService} in order to request
@@ -684,10 +671,23 @@ define([],function() {
        * @since 1.0.39
        *
        * @event
-       * @param {string} [action] The action to perform
-       * @param {string} [actionTopic] The topic to re-publish the action request on
+       * @property {string} [action] The action to perform
+       * @property {string} [actionTopic] The topic to re-publish the action request on
        */
       SELECTED_DOCUMENTS_ACTION: "ALF_SELECTED_DOCUMENTS_ACTION_REQUEST",
+
+      /**
+       * Used to indicate the list of currently selected documents has changed and provides the details of those items.
+       * 
+       * @instance
+       * @type {string} 
+       * @default
+       * @since 1.0.35
+       * 
+       * @event module:alfresco/core/topics~SELECTED_DOCUMENTS_CHANGED
+       * @property {object[]} selectedItems - The items that are selected
+       */
+      SELECTED_DOCUMENTS_CHANGED: "ALF_SELECTED_FILES_CHANGED",
 
       /**
        * This topic can be published to set a user preference. It is typically handled by 
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -673,6 +673,18 @@ define([],function() {
       SELECTED_DOCUMENTS_CHANGED: "ALF_SELECTED_FILES_CHANGED",
 
       /**
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       *
+       * @event
+       * 
+       */
+      SELECTED_DOCUMENTS_ACTION: "ALF_SELECTED_DOCUMENTS_ACTION_REQUEST",
+
+      /**
        * This topic can be published to set a user preference. It is typically handled by 
        * the [PreferenceService]{@link module:alfresco/services/PreferenceService}.
        *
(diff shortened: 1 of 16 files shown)


aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -673,6 +673,23 @@ define([],function() {
...
+       * be forwarded onto the [ActionService]{@link module:alfresco/services/ActionService} in order to request
+       * that the configured action be performed on the currently selected items.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       *
+       * @event
+       * @param {string} [action] The action to perform
```
MartinDoyleUK: @property ...

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -673,6 +673,23 @@ define([],function() {
...
+       * @param {string} [actionTopic] The topic to re-publish the action request on
+       */
+      SELECTED_DOCUMENTS_ACTION: "ALF_SELECTED_DOCUMENTS_ACTION_REQUEST",
```
MartinDoyleUK: Should be before `SELECTED_DOCUMENTS_CHANGED` (alphabetical)

aikau/src/main/resources/alfresco/core/topics.js
MartinDoyleUK: This should be a function, I believe, not a string? (cf. line 275 of `ActionService`)

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -673,6 +673,23 @@ define([],function() {
...
+       * that the configured action be performed on the currently selected items.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.39
+       *
+       * @event
+       * @param {string} [action] The action to perform
+       * @param {string} [actionTopic] The topic to re-publish the action request on
```
MartinDoyleUK: Might be good to mention that one or other is required in order for anything to happen (although that might be so obvious as to be redundant)

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
```diff
@@ -0,0 +1,261 @@
...
+   return declare([Core, _AlfDocumentListTopicMixin], {
+      
+      /**
+       * This is used to keep track of the documents that are currently selected. It is initialised to an empty
+       * array in the constructor, [onItemSelected]{@link module:alfresco/lists/SelectedItemStateMixin#onItemSelected} 
+       * adds elements and the [onItemDeselected]{@link module:alfresco/lists/SelectedItemStateMixin#onItemDeselected} 
+       * removes them.
+       *
+       * @instance
+       * @type {object}
```
MartinDoyleUK: `@type {object[]}`

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
```diff
@@ -0,0 +1,261 @@
...
+       * An array of the [itemKeyProperty]{@link module:alfresco/lists/SelectedItemStateMixin#itemKeyProperty}
+       * attributes taken from each entry in the 
+       * [currentlySelectedItems]{@link module:alfresco/lists/SelectedItemStateMixin#currentlySelectedItems}
+       * map.
+       * 
+       * @instance
+       * @type {string[]}
+       * @default
+       */
+      selectedItems: null,
```
MartinDoyleUK: Very similar in name to the `currentlySelectedItems` property, which is an array of objects, yet this is an array of strings. Could this be called something like `selectedItemKeys`?

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
```diff
@@ -0,0 +1,261 @@
...
+
+      /**
+       * This is used to keep a reference to a timeout that is started on the publication of a selected document
+       * topic. It is important that multiple selection events can be captured so that only one publication of
+       * selected items occurs.
+       *
+       * @instance
+       * @type {timeout}
```
MartinDoyleUK: Should this be `number` or `object` instead? `timeout` isn't really a type, just a description of the variable.

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
```diff
@@ -0,0 +1,261 @@
...
+       * This function is called if 
+       * [disableWhenNothingSelected]{@link module:alfresco/lists/SelectedItemStateMixin#disableWhenNothingSelected}
+       * is configured to be true and a selection update occurs that results in no items being selected. By
+       * default it will attempt to toggle the disable state of the widget.
+       *
+       * @instance
+       * @param {boolean} disable Indicates whether to disable or enable the widget
+       * @overridable
+       */
+      disable: function alfresco_lists_SelectedItemStateMixin__disable(disable) {
```
MartinDoyleUK: `disable(false)` is less readable than `setDisabled(false)`, subjectively ...

aikau/src/main/resources/alfresco/renderers/Selector.js
```diff
@@ -144,7 +144,8 @@ define(["dojo/_base/declare",
                   var b = lang.getObject(this.itemKey, false, this.currentItem);
                   var match = ((a || a === 0) && a === b);
                   if (match) {
-                     this.select();
+                     domClass.add(this.selectorNode, "checked");
+                     domClass.remove(this.selectorNode, "unchecked");
```
MartinDoyleUK: Is this existing code? Is it easy to change to have one single "boolean class", rather than two that must not both be present?

aikau/src/main/resources/alfresco/services/ActionService.js
```diff
@@ -385,7 +385,7 @@ define(["dojo/_base/declare",
        *
        * @instance
        */
-      onSelectedFilesChanged: function alfresco_services_ActionService__onSelectedFilesChanged() {
+      onselectedItemsChanged: function alfresco_services_ActionService__onselectedItemsChanged() {
```
MartinDoyleUK: Is there a reason for this having a lower-case `s`? Am assuming there is, as it looks deliberate.

aikau/src/main/resources/webscript-libs/doclib/doclib.lib.js
```diff
@@ -26,7 +26,7 @@ function getSyncMode() {
 function getUserDocLibPreferences() {
 
    // Initialise some default preferences...
-   var docLibPrefrences = {
+   var docLibPreferences = {
```
MartinDoyleUK: Pedanticus approves.

aikau/src/main/resources/alfresco/core/topics.js
draperd: No, it's still a string that's included in the payload.

aikau/src/main/resources/alfresco/core/topics.js
draperd: Yes, that's probably a redundant statement to add.

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
draperd: It's not an array, it's an object. See code where it's instantiated.

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
draperd: currentlySelectedItems is an *object*.... selectedItems is an *array of objects*

aikau/src/main/resources/alfresco/renderers/Selector.js
draperd: It is existing class and it's used elsewhere. The whole thing could do with some rework, all of which was out of scope of this issue.

aikau/src/main/resources/alfresco/services/ActionService.js
draperd: Nope, I'll change it

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
MartinDoyleUK: Then you need to correct the JSDoc comment just above that says it's initialised to an empty array! :)

aikau/src/main/resources/alfresco/lists/SelectedItemStateMixin.js
MartinDoyleUK: Then why is it listed as an array of strings? The naming comment remains, if it is indeed an array of strings.

aikau/src/main/resources/alfresco/core/topics.js
MartinDoyleUK: Then why is there an ultimate `typeof==="function"` test against it?

draperd: You mean the 
```
if (typeof this[payload.action] === "function") 
```
line in action service? It's the *this* bit that is key

MartinDoyleUK: Sorry. What a dumbass. You're right ... please ignore me :)

PR #608, opened 2015-10-14
AKU-627: Thumbnail select on click

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-627 to add optional support for item selection by clicking on it's thumbnail. A new mixin module ItemSelectionMixin has been abstracted from the Selector widget and mixed into both the Selector and Thumbnail widgets. The unit tests have been updated to verify the new functionality works as expected.

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
@@ -178,6 +178,9 @@ define(["dojo/_base/declare",
        * @instance
        * @type {string} 
        * @default
+       * @event
+       * @property {string} [value] One of "selectAll", "selectNone", "selectInvert", "selectFolders" or "selectDocuments"
+       * @property {object[]} [selectedItems] An array of items that have been selected
        */
       documentSelectionTopic: topics.DOCUMENT_SELECTION_UPDATE,
       
@@ -187,6 +190,8 @@ define(["dojo/_base/declare",
        * @instance
        * @type {string} 
        * @default
+       * @event
+       * @property {object} value The item selected
        */
       documentSelectedTopic: topics.DOCUMENT_SELECTED,
       
@@ -205,6 +210,8 @@ define(["dojo/_base/declare",
        * @instance
        * @type {string} 
        * @default
+       * @event
+       * @property {object} value The item de-selected
        */
       documentDeselectedTopic: topics.DOCUMENT_DESELECTED,
       
--- a/aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
@@ -177,10 +177,8 @@ define(["dojo/_base/declare",
        * 
        * @instance
        * @type {string} 
-       * @default
+       * @default [DOCUMENT_SELECTION_UPDATE]{@link module:alfresco/core/topics#DOCUMENT_SELECTION_UPDATE}
        * @event
-       * @property {string} [value] One of "selectAll", "selectNone", "selectInvert", "selectFolders" or "selectDocuments"
-       * @property {object[]} [selectedItems] An array of items that have been selected
        */
       documentSelectionTopic: topics.DOCUMENT_SELECTION_UPDATE,
       
--- a/aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
+++ b/aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
@@ -1,0 +1,252 @@
+/**
+ * Copyright (C) 2005-2015 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * This module should be mixed into renderers that wish to be able to either display selected state or that
+ * can be clicked to change selected state. This is mixed into both the [Selector]{@link module:alfresco/renderers/Selector}
+ * and [Thumbnail]{@link module:alfresco/renderers/Thumbnail} renderers.
+ * 
+ * @module alfresco/lists/ItemSelectionMixin
+ * @author Dave Draper
+ * @since 1.0.40
+ */
+define(["dojo/_base/declare",
+        "alfresco/core/Core",
+        "alfresco/core/topics",
+        "alfresco/documentlibrary/_AlfDocumentListTopicMixin",
+        "dojo/_base/lang",
+        "dojo/_base/array",
+        "dojo/dom-class",
+        "dojo/_base/event"], 
+        function( declare, Core, topics, _AlfDocumentListTopicMixin, lang, array, domClass, Event) {
+   
+   return declare([Core, _AlfDocumentListTopicMixin], {
+      
+      /**
+       * The dot-notation property in the currentItem that uniquely idenfities that item. This defaults
+       * to "nodeRef" as the most likely use case is for working with Nodes but this can be configured
+       * to a different value.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       */
+      itemKey: "nodeRef",
+
+      /**
+       * This indicates whether or not the mixing widget should subscribe to item selection events. This
+       * attribute is inspected by the [createItemSelectionSubscriptions]{@link module:alfresco/lists/ItemSelectionMixin#createItemSelectionSubscriptions}
+       * function to determine whether to bind the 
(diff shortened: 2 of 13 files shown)


aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
```diff
@@ -178,6 +178,9 @@ define(["dojo/_base/declare",
        * @instance
        * @type {string} 
        * @default
+       * @event
+       * @property {string} [value] One of "selectAll", "selectNone", "selectInvert", "selectFolders" or "selectDocuments"
+       * @property {object[]} [selectedItems] An array of items that have been selected
```
MartinDoyleUK: I thought we were going to use the @default tag to link people to the documentation that describes the event better (i.e. no default value and I think the property tags are redundant?)

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
```diff
@@ -0,0 +1,252 @@
...
+      /**
+       * The dot-notation property in the currentItem that uniquely idenfities that item. This defaults
+       * to "nodeRef" as the most likely use case is for working with Nodes but this can be configured
+       * to a different value.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       */
+      itemKey: "nodeRef",
```
MartinDoyleUK: Is there any mileage in having this as `@type {string|string[]}` so that people could also specify a preference order of properties, e.g. `["nodeRef", "node.nodeRef"]`?

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
```diff
@@ -0,0 +1,252 @@
...
+      /**
+       * Returns whether or not selection subscriptions and publications should be made globally. By default
+       * this returns [publishGlobal]{@link module:alfresco/renderers/_PublishPayloadMixin#publishGlobal}.
+       *
+       * @instance
+       * @overridable
+       * @return {boolean} A boolean indicating whether or not to publish and subscribe to selection topics globally.
+       */
+      getSelectionPublishGlobal: function alfresco_lists_ItemSelectionMixin__getSelectionPublishGlobal() {
+         return this.publishGlobal;
```
MartinDoyleUK: This variable does not exist, as far as I can tell, in the prototype chain. We need to either mixin another module or to include this in the defined variables.

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
```diff
@@ -0,0 +1,252 @@
...
+      /**
+       * Returns whether or not selection subscriptions and publications should be made globally. By default
+       * this returns [publishToParent]{@link module:alfresco/renderers/_PublishPayloadMixin#publishToParent}.
+       *
+       * @instance
+       * @overridable
+       * @return {boolean}  A boolean indicating whether or not to publish and subscribe to selection topics to the parent scope.
+       */
+      getSelectionPublishToParent: function alfresco_lists_ItemSelectionMixin__getSelectionPublishToParent() {
+         return this.publishToParent;
```
MartinDoyleUK: As per comment above (mixin or define)

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
```diff
@@ -0,0 +1,252 @@
...
+      /**
+       * Updates the CSS classes to indicate that the item has been selected and publishes 
+       * [the topic]{@link module:alfresco/documentlibrary/_AlfDocumentListTopicMixin#documentSelectedTopic}
+       * to indicate that an item has been selected.
+       * 
+       * @instance
+       * @fires module:alfresco/documentlibrary/_AlfDocumentListTopicMixin#documentSelectedTopic
+       */
+      select: function alfresco_lists_ItemSelectionMixin__select() {
+         domClass.add(this.getSelectorNode(), "alfresco-lists-ItemSelectionMixin--checked");
```
MartinDoyleUK: This needs to be more semantic, and state should be applied as high as possible, in case other elements need to adjust their styling because of it.

```javascript
domClass.add(this.domNode, "alfresco-lists-ItemSelectionMixin--selected");
```

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
```diff
@@ -0,0 +1,252 @@
...
+ */
+define(["dojo/_base/declare",
+        "alfresco/core/Core",
+        "alfresco/core/topics",
+        "alfresco/documentlibrary/_AlfDocumentListTopicMixin",
+        "dojo/_base/lang",
+        "dojo/_base/array",
+        "dojo/dom-class",
+        "dojo/_base/event"], 
+        function( declare, Core, topics, _AlfDocumentListTopicMixin, lang, array, domClass, Event) {
```
MartinDoyleUK: Why the capital E?

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
```diff
@@ -0,0 +1,252 @@
...
+       * or [select]{@link module:alfresco/lists/ItemSelectionMixin#select} (depending upon the current
+       * state).
+       * 
+       * @instance
+       */
+      onSelectionClick: function alfresco_lists_ItemSelectionMixin__onSelectionClick(evt) {
+         if (this.selectOnClick)
+         {
+            evt && Event.stop(evt);
+            if (domClass.contains(this.getSelectorNode(), "alfresco-lists-ItemSelectionMixin--checked"))
```
MartinDoyleUK: Could just have a single function (`select()` or `setSelected()`) that takes a boolean? Could then write this as:

```javascript
this.setSelected(domClass.contains(this.domNode, "alfresco-lists-ItemSelectionMixin--selected"));
```

Also, this string literal that's repeated three-times should really be a variable?

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -122,6 +150,115 @@ define(["dojo/_base/declare",
...
+       * and "folder-256.png". The default is "folder-64.png".
+       *
+       * @instance
+       * @type {string}
+       * @default
+       */
+      folderImage: "folder-64.png",
+
+      /**
+       * The property to use for the image id. Defaults to "jsNode.nodeRef.nodeRef"
```
MartinDoyleUK: We don't need to put in the comment what it defaults to, as it is (a) written just below this in the JSDoc because of the @default tag and (b) it can rust

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -122,6 +150,115 @@ define(["dojo/_base/declare",
...
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       */
+      imageIdProperty: "jsNode.nodeRef.nodeRef",
+
+      /**
+       * The property to use for the image title. Defaults to "displayName"
```
MartinDoyleUK: As above regarding @default tag

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -408,31 +512,40 @@ define(["dojo/_base/declare",
...
+         // Check whether or not the thumbnail should ONLY support selection on click (because in 
+         // some circumstances, e.g. film strip view carousel we might want to select AND perform
+         // an additinal action)...
+         if (!this.onlySelectOnClick)
          {
-            this.nodePromise = this.currentItem;
-            var type = lang.getObject("node.type", false, this.currentItem),
-                mimetype = lang.getObject("node.mimetype", false, this.currentItem);
-            if (!type || !mimetype)
+            // TODO: Need to use a nodeRef property attribute that can be configured
```
MartinDoyleUK: Are we not doing that now??

aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
draperd: Can you add what you want to go here to save the to-ing and fro-ing :)

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
draperd: Because that's how it was from the module I took it from, which was probably based on a sample I took from the Dojo documentation.... seriously, does it matter? ;)

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
draperd: Maybe, but it's out of scope for this issue.

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
draperd: No, because they existed previously in Selector (which now mixes in this module). If someone has extended Selector they maybe calling, overriding or extending select and/or deslect .

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
draperd: No, it means that we could be using something like itemKeyProperty, but with the inconsistent REST APIs this is quite tricky to do. For backwards compatibility we'd need to check the itemKeyProperty and then "nodeRef" and "node.nodeRef"... it would all get a bit messy and is also out of scope for this issue.

aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
MartinDoyleUK: Populate the @default value to be the topics.XXX event
Keep the @event tag
Lose the @property tags

aikau/src/main/resources/alfresco/lists/ItemSelectionMixin.js
MartinDoyleUK: The code will run fine :P

PR #609, opened 2015-10-14
Add form field focus and submit on enter

spotnape:
Add optional functionality to focus first form field and submit by pressing the enter key. Docs, tests and reformatting included.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/Form.js
+++ b/aikau/src/main/resources/alfresco/forms/Form.js
@@ -117,9 +117,10 @@ define(["dojo/_base/declare",
         "dijit/registry",
         "dojo/Deferred",
         "dojo/dom-construct",
-        "dojo/dom-class"], 
+        "dojo/dom-class",
+        "dojo/on"], 
         function(declare, _Widget, _Templated, Form, AlfCore, CoreWidgetProcessing, topics, _AlfHashMixin, RulesEngineMixin, 
-                 template, ioQuery, Warning, hashUtils, lang, AlfButton, array, registry, Deferred, domConstruct, domClass) {
+                 template, ioQuery, Warning, hashUtils, lang, AlfButton, array, registry, Deferred, domConstruct, domClass, on) {
    
    return declare([_Widget, _Templated, AlfCore, CoreWidgetProcessing, _AlfHashMixin, RulesEngineMixin], {
       
@@ -309,7 +310,27 @@ define(["dojo/_base/declare",
        * @since 1.0.32
        */
       warningsPosition: "top",
-      
+
+      /**
+       * Should the first field in the form be focused when the form is loaded?
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.40
+       */
+      firstFieldFocusOnLoad: false,
+
+      /**
+       * Should the form automatically submit when the enter key is pressed?
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.40
+       */
+      submitOnEnter: false,
+
       /**
        * @instance
        */
@@ -383,6 +404,24 @@ define(["dojo/_base/declare",
             }, this);
 
             this.processWidgets(this.widgets, this._form.domNode, "FIELDS");
+         }
+
+         // If configured, focus the first field in the form
+         if (this.firstFieldFocusOnLoad)
+         {
+            this.focus();
+         }
+
+         // If configured, allow the form to be submitted by pressing the enter key
+         if (this.submitOnEnter && !this.autoSavePublishTopic)
+         {
+            on(this, "keydown", lang.hitch(this, function(event){
+               if (event.keyCode === dojo.keys.ENTER && this.invalidFormControls.length === 0)
+               {
+                  var payload = lang.mixin(this.okButtonPublishPayload || {}, this.getValue());
+                  this.alfPublish(this.okButtonPublishTopic, payload, (this.okButtonPublishGlobal !== undefined && this.okButtonPublishGlobal === true));
+               }  
+            }));
          }
       },
 
--- a/aikau/src/test/resources/alfresco/forms/controls/BaseFormTest.js
+++ b/aikau/src/test/resources/alfresco/forms/controls/BaseFormTest.js
@@ -29,146 +29,193 @@ define(["intern!object",
         "alfresco/TestCommon"], 
         function(registerSuite, assert, keys, TestCommon) {
 
-registerSuite(function(){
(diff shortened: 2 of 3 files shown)


draperd:
The problem with this PR is that it only addresses a single use case... I've had preliminary conversations with the UX team about the broader behaviour of how enter key presses impact form (e.g. navigation to next field if more required data is pending, only submit when all required fields have data provided, etc.)

There are larger things to consider than just a single narrow use case. I'm not going to merge this PR, I want us to have a proper review of all the implications of such as a change, as it is there are issues with the requested change anyway (e.g. not backwards compatible, tests not covering a broad range of widgets that may already include enter key press handlers). 

I would have preferred this to have been discussed before the effort went into creating the PR. In this instance we have evaluated the broader issues it might be better to create a custom widget or custom form.

aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -309,7 +310,27 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * Should the first field in the form be focused when the form is loaded?
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.40
+       */
+      firstFieldFocusOnLoad: false,
```
draperd: Not backwards compatible - the first field currently is focused on form load.

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/forms/controls/BaseForm.get.js
```diff
@@ -172,6 +172,34 @@ model.jsonModel = {
...
+                  id: "FOCUS_AND_SUBMIT_ON_ENTER_FORM",
+                  config: {
+                     okButtonPublishTopic: "SUBMIT_FOCUS_AND_SUBMIT_ON_ENTER_FORM",
+                     okButtonPublishGlobal: true,
+                     firstFieldFocusOnLoad: true,
+                     submitOnEnter: true,
+                     widgets: [
+                        {
+                           id: "FORM_FIELD1",
+                           name: "alfresco/forms/controls/TextBox",
```
draperd: We should be testing a range of widgets here - not just text box, in particular we need consideration over what happens with widgets that have their own keypress listeners.

aikau/src/test/resources/alfresco/forms/controls/BaseFormTest.js
```diff
@@ -29,146 +29,193 @@ define(["intern!object",
...
+            .findByCssSelector("#FOCUS_AND_SUBMIT_ON_ENTER_FORM #FORM_FIELD2 div.dijitFocused")
+               .then(function () {
+                  assert.fail(null, null, "Form field 2 should not be focused");
+               },
+               function () {
+                  
+               });
+         },
+
+         "Check the form submits by the enter key": function() {
```
draperd: There is no test that a form is NOT submitted when the new configuration value isn't applied.

aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -384,6 +405,24 @@ define(["dojo/_base/declare",
...
+         if (this.firstFieldFocusOnLoad)
+         {
+            this.focus();
+         }
+
+         // If configured, allow the form to be submitted by pressing the enter key
+         if (this.submitOnEnter && !this.autoSavePublishTopic)
+         {
+            on(this, "keydown", lang.hitch(this, function(event){
+               if (event.keyCode === dojo.keys.ENTER && this.invalidFormControls.length === 0)
```
draperd: Should use AMD dependency here rather than referencing dojo.keys

aikau/src/main/resources/alfresco/forms/Form.js
draperd: We're now trying to ensure that variables are listed alphabetically rather than randomly (yes, this isn't applied throughout the code base entirely, but you'll see that with each PR we are making updates the code).

aikau/src/test/resources/alfresco/forms/controls/BaseFormTest.js
```diff
@@ -29,146 +29,193 @@ define(["intern!object",
...
+   // This test reloads the page to clear any previous focus
+   registerSuite(function(){
+      var browser;
+
+      return {
+         name: "Base Form Control Tests - Focus and Enter to Submit",
+
+         setup: function() {
+            browser = this.remote;
+            return TestCommon.loadTestWebScript(this.remote, "/BaseForm", "Base Form Control Tests").end();
```
draperd: I would have preferred an individual test page and test suite for this capability (see some of the other form tests) this helps when browsing the available pages from the unit test index page.

draperd:
I've tried to make some useful comments on what to consider for future PRs, these aren't suggestions to be implemented now on this (closed) PR... we need to have the broader discussion to work out how to take this further.

aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -384,6 +405,24 @@ define(["dojo/_base/declare",
...
+               {
+                  var payload = lang.mixin(this.okButtonPublishPayload || {}, this.getValue());
```
draperd: This would be better as shared code, e.g. maybe actioning the confirmation button directly or getting it's payload rather than duplicating code to create the payload.

aikau/src/main/resources/alfresco/forms/Form.js
draperd: What happens if there is more than one form on the page and that both request firstFieldFocusOnLoad to be true? Or there is another widget that also requests focus. This configuration attribute is making a promise that is can't necessarily keep.

draperd:
On top of everything else, I'm not even sure this is fully addressing the very specific use case you have... e.g. user arrives on login page, first field is focused, types user name hits enter - nothing happens, I would have thought from a pure user flow you'd definitely want to navigate to the next field on enter and not force the user to use tab.

MartinDoyleUK:
On that very final comment about enter moving to next field on login form, the consistency of always knowing that ENTER does a single job (submit form) and TAB always does a single job (move to next field) does weigh heavily against your argument.

draperd:
That's because you know what TAB does. I suspect that your more basic user never touches the TAB key.

MartinDoyleUK:
If someone has used the keyboard to navigate even once before, in any application, on any operating system, they will know what TAB does. You cannot play around with such ubiquitous precedents without an incredibly strong reason.

PR #611, opened 2015-10-15
AKU-634: Non-resizeable sidebar

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-634 to update the AlfSideBarContainer so that it can be configured such that the sidebar cannot be resized. The CSS has also been updated to improve the use of LESS variables to make it possible to theme the background colours of the sidebar, the main panel and the borders used on the separators.


aikau/src/main/resources/alfresco/layout/AlfSideBarContainer.js
```diff
@@ -128,6 +155,16 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * Indicates whether or not the sidebar should be resizable or not.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.40
+       */
+      isResizeable: true,
```
MartinDoyleUK: The rest of Aikau uses "resizable" without the extra "e", as in the comment for this variable, so I think we should be consistent here and drop the extra "e".

PR #615, opened 2015-10-16
Create 'color-two' AlfButton class

spotnape:
...from accessible contrast palette (green colour in defaults.less)


draperd:
Can you provide a bit more information on what this is all about? Is there a related JIRA ticket with screenshots, etc.?

I'm not sure about the names as they're not very explanatory - nor are they particularly consistent, e.g. you're adding a color 2 - but there is no colour 1 !

Also, the selectors aren't very descriptive - whereas call-to-action is somewhat (if not entirely) explanatory it's not clear what the purpose of the second button colour is.

spotnape:
Please refer to this JIRA for an example of this (green) button in use on the left side: https://issues.alfresco.com/jira/browse/SFS-31

I didn't want to use the phrase green because there is no guarantee that it will be green - it is up to the theme to define that. In a high contrast theme it could be yellow on black. Neither will this colour necessarily be fixed to this one function - it is just the second colour in the accessible palette that UX have produced. Unfortunately color-one IS the call-to-action button which is why I started at two.

draperd:
I'm a bit confused... there are two buttons on those screenshots, the green would be the call-to-action and the white one would be a regular button. Where is the requirement for a third button in those screenshots?

If we need a third class of button I'd prefer to give it a semantically meaningful name.

spotnape:
I believe that a theme should be applied across an application. We already have a requirement for a regular call-to-action button on the login page for example. The colour of that button, for the standard theme, matches the default call-to-action button provided by Aikau. Therefore, to provide the button as shown in the example above, which does not match the default call-to-action button, I need a third button colour scheme.

draperd:
What's the difference between the login button and the upload button? Are they not examples of call-to-action (e.g. this is the button that draws the eye)... my question would be why the designs are inconsistent?

Are call-to-action buttons going to be different on every page or just on login page? Is the button upload a special case? What is the general UX policy going to be for buttons - are there different colours for different classes for actions?

Does authentication (e.g. login) have a different colour than creation (e.g. upload)?

PR #616, opened 2015-10-16
AKU-631: Update Thumbnails for aspect ratio, sizing and cropping

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-631 to work through the various thumbnail widgets and generally improve their handling of sizing, cropping and aspect ratio. I've also added in configurable support for styling and pushed default styling in LESS variables. 

I've intentionally removed the DND highlighting and link overlays (because these are not used in production yet and we know that there are backlog items on the horizon to address these).


draperd:
Don't merge this yet ... I have to push some JSDoc updates and add some more tests, but I thought it might be worth you having a preliminary review as there are a lot of changes.

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -188,6 +263,40 @@ define(["dojo/_base/declare",
...
+       * to be false then portrait thumbnails will be clipped at the bottom and landscape thumbnails will be 
+       * vertically stretched. However, if no [dimensions]{@link module:alfresco/renderers/Thumbnail#dimensions}
+       * are provided then the aspect ration will always be maintained.
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.40
+       */
+      maintainAspectRatio: true,
```
MartinDoyleUK: Got some problems with the use of "aspect ratio" in general. Changing the aspect ratio of an image certainly could be applied to cropping it, but it more normally refers to stretching it. I'd prefer to see terms such as fill/crop and fit/letterbox, which more clearly describe what's happening with the image, which is not actually changing, only the window through which it's being viewed. Realise this is subjective, but don't want to say nothing when it makes me feel so uncomfortable as-is.

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -188,6 +263,40 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * This will be set to the natural height of the displayed image.
+       *
+       * @instance
+       * @type {number}
+       * @default
+       * @since 1.0.40
+       */
+      naturalImageHeight: null,
```
MartinDoyleUK: Maybe a @readonly tag?

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -188,6 +263,40 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * This will be set to the natural width of the displayed image.
+       *
+       * @instance
+       * @type {number}
+       * @default
+       * @since 1.0.40
+       */
+      naturalImageWidth: null,
```
MartinDoyleUK: Maybe a @readonly tag?

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -499,6 +619,161 @@ define(["dojo/_base/declare",
...
+       */
+      cropImage: function alfresco_renderers_Thumbnail__cropImage() {
+         domStyle.set(this.frameNode, {
+            position: "relative",
+            overflow: "hidden",
+            textAlign: "left"
+         });
+         if (this.naturalImageHeight > this.naturalImageWidth)
+         {
+            var vOffset = ((this.imageNodeWidth / (this.naturalImageWidth / this.naturalImageHeight)) - this.imageNodeWidth) / 2;
```
MartinDoyleUK: Might be nice to have `this.naturalImageWidth / this.naturalImageHeight` as a `naturalAspectRatio` variable, and then add an additional `isPortrait = naturalAspectRatio < 1` variable, making the condition above easier to read (i.e. `if (isPortrait) {`) as well as reducing duplicated calculations in the offset calculations.

aikau/src/main/resources/alfresco/renderers/css/Thumbnail.css
```diff
@@ -1,48 +1,52 @@
...
+      text-align: center;
+   }
 
-.alfresco-renderers-Thumbnail span.link.hidden
-{
-   display: none;
-}
+   .alfresco-renderers-Thumbnail__image {
+      cursor: pointer;
+      max-height: ~'calc(100% - @thumbnail-margin - @thumbnail-margin - @thumbnail-border-width)';
```
MartinDoyleUK: I can't remember why we allow `calc` when it doesn't work in IE8? I thought it was because it didn't matter when we'd previously used it, but this would presumably look bad in IE8?? There may be another reason entirely though.

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -160,6 +194,27 @@ define(["dojo/_base/declare",
...
 
       /**
+       * Indicates whether or not the thumbnail image should be given a shadow effect.
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.40
+       */
+      hasShadow: false,
```
MartinDoyleUK: This feels like CSS not a JS property. Can't we have a default that looks good and let people style this as necessary? Doesn't feel right having it as config, but I guess if there's a specific need to have it ... ?

MartinDoyleUK:
Good dog photos :)

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
draperd: Yes.... the good reason is that this is a declarative framework and we need to give developers options without resorting to writing custom CSS and/or custom widgets.

aikau/src/main/resources/alfresco/renderers/css/Thumbnail.css
draperd: It's the difference between "looks bad on IE8" and "fundamentally doesn't work on IE8".... we can live with looking bad (after all, that'll probably be the least of the users problems).

Having said that, looking at the code I wonder if we can do the calculation entirely in LESS? After all the 100% here is simply going to be the thumbnail height. I'll look at that again.

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
draperd: The offset calculations aren't duplicated - you couldn't re-use an aspect variable (one is width over height, the other is height over width). To ease readability though I'll improve the commenting around the code rather than adding in additional instance variables that won't be used elsewhere

PR #618, opened 2015-10-20
AKU-628: Resize thumbnail view

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-628 to add support for resizing thumbnails directly (rather than indirectly by adjusting the number of columns).

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -690,6 +690,21 @@ define([],function() {
       SELECTED_DOCUMENTS_CHANGED: "ALF_SELECTED_FILES_CHANGED",
 
       /**
+       * Use by the [AlfGalleryViewSlider]{@link module:alfresco/documentlibrary/AlfGalleryViewSlider}
+       * to publish changes in the number of columns that should be shown in the 
+       * [AlfGalleryView]{@link module:alfresco/documentlibrary/views/AlfGalleryView}.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.40
+       *
+       * @event
+       * @property {number} value - The number of columns to display.
+       */
+      SET_COLUMNS: "ALF_DOCLIST_SET_GALLERY_COLUMNS",
+
+      /**
        * This topic can be published to set a user preference. It is typically handled by 
        * the [PreferenceService]{@link module:alfresco/services/PreferenceService}.
        *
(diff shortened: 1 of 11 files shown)


aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -690,6 +690,21 @@ define([],function() {
...
+       * to publish changes in the number of columns that should be shown in the 
+       * [AlfGalleryView]{@link module:alfresco/documentlibrary/views/AlfGalleryView}.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.40
+       *
+       * @event
+       * @property {number} value - The number of columns to display.
```
MartinDoyleUK: The hyphen isn't necessary here (or in the many other property tags in this file with hyphens), however I've checked the generated documentation and it doesn't seem to affect it. I would suggest though we try to be consistent, and we don't use hyphens anywhere else (or at least don't need to, and normally don't), so would suggest we just quickly strip out the hyphens in this file at some point. I don't think it should stop this PR being merged though.

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -701,6 +716,21 @@ define([],function() {
       SET_PREFERENCE: "ALF_PREFERENCE_SET",
 
       /**
+       * Use by the [AlfGalleryViewSlider]{@link module:alfresco/documentlibrary/ThumbnailSizeSlider}
```
MartinDoyleUK: Copy/paste error?

aikau/src/main/resources/alfresco/documentlibrary/ThumbnailSizeSlider.js
```diff
@@ -0,0 +1,177 @@
...
+      /**
+       * The number of steps (including the beginning and end positions) on the slider
+       * @instance
+       * @type {integer}
+       * @default
+       */
+      discreteValues: 16,
+         
+      /**
+       * The minimum value on the slider (this will represent the smalled size in pixels that a thumbnail
```
MartinDoyleUK: *smallest

aikau/src/main/resources/alfresco/documentlibrary/views/AlfGalleryView.js
```diff
@@ -60,6 +83,75 @@ define(["dojo/_base/declare",
...
+       * @instance
+       * @type {string}
+       * @default
+       */
+      nextLinkLabel: null,
+
+      /**
+       * The topic to publish when the next link is clicked.
+       * 
+       * @event nextLinkPublishTopic
```
MartinDoyleUK: just `@event` is necessary

PR #620, opened 2015-10-20
AKU-632: Clickable action icon

MartinDoyleUK:
This PR addresses issue [AKU-632](https://issues.alfresco.com/jira/browse/AKU-643) by creating a new Image widget and also updates the Logo widget to be as light as possible by extending the new Image class.

__NOTE: Please don't merge, as still have yet to create unit-tests. This pull-request is for early-access review of the architecture of the changes.__


aikau/src/main/resources/alfresco/html/Image.js
```diff
@@ -0,0 +1,421 @@
...
+
+/**
+ * This module provides a simple mechanism for creating image elements on a page. Additionally, it mixes
+ * in the [_PublishOrLinkMixin]{@link module:alfresco/core/_PublishOrLinkMixin} class to provide support
+ * for clicking to navigate or to publish a topic.
+ *
+ * @module alfresco/html/Image
+ * @extends external:dijit/_WidgetBase
+ * @mixes external:dojo/_TemplatedMixin
+ * @author Martin Doyle
```
draperd: Missing @since

aikau/src/main/resources/alfresco/enums/urlTypes.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * <p>This enum contains all possible values of URL type.</p>
+ * 
+ * @module alfresco/enums/urlTypes
+ * @author Martin Doyle
```
draperd: Missing @since

aikau/src/main/resources/alfresco/html/Image.js
draperd: Missing @mixes PublishOrLinkMixin

aikau/src/main/resources/alfresco/html/Image.js
```diff
@@ -0,0 +1,421 @@
...
+         defaultMinimumSize: 50,
+
+         /**
+          * An optional CSS height to apply to the image node (e.g. "10px" or "50%")
+          *
+          * @instance
+          * @type {string}
+          * @default
+          */
+         height: null,
```
draperd: If we're considering eventually making Thumbnail extend this module then we want to make sure that we use consistent variable names. So Thumbnail just uses "width" as a shorthand for both "width" and "height" and then "dimensions" for the longhand form. We want to avoid inconsistencies in naming, Thumbnail is already out in the wild (and has been for a while) so can't be changed.

aikau/src/main/resources/alfresco/html/Image.js
```diff
@@ -0,0 +1,421 @@
...
+
+         /**
+          * If we can get no other sizes, then rather than having a zero-by-zero image,
+          * make it square using this value (treated as pixels).
+          *
+          * @instance
+          * @type {number}
+          * @default
+          */
+         defaultMinimumSize: 50,
```
draperd: Why not just have a default width?

aikau/src/main/resources/alfresco/html/Image.js
```diff
@@ -0,0 +1,421 @@
...
+         naturalWidth: 0,
+
+         /**
+          * The path to the default blank gif, using the module pathing system
+          *
+          * @instance
+          * @type {string}
+          * @default
+          */
+         pathToBlankGif: "alfresco/css/images/blank.gif",
```
draperd: Should this be a @readonly ?

aikau/src/main/resources/alfresco/html/Image.js
```diff
@@ -0,0 +1,421 @@
...
+            };
+         },
+
+         /**
+          * Get the natural size of the currently displayed image.
+          *
+          * @instance
+          * @returns {external:dojo/promise/Promise} A promise that will resolve to a [dimensions]{@link module:alfresco/html/Image#Dimensions} object
+          */
+         getNaturalDimensions: function alfresco_html_Image__getNaturalDimensions() {
```
draperd: As with earlier comment... if future alignment with Thumbnail is planned, this should be getNaturalImageSize

aikau/src/main/resources/alfresco/html/Image.js
```diff
@@ -0,0 +1,421 @@
...
+         },
+
+         /**
+          * Get the natural size of the currently displayed image.
+          *
+          * @instance
+          * @param {boolean} [force] Whether to force retrieval if already present
+          * @returns {external:dojo/promise/Promise} A promise that will resolve once dimensions are retrieved
+          */
+         _updateNaturalDimensions: function alfresco_html_Image___updateNaturalDimensions(force) {
```
draperd: Is it not possible to do this the same way as Thumbnail? Rather than creating and throwing away a new image element.

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: Width is width and height? That's literally insane. I understand backward compatibility, but surely not at the expense of all reason? We can certainly add dimensions support, for providing both at once, and we can add a property that will default width to height and vice-versa if only one is provided, but to have that as default is nuts.

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: Because there are two dimensions on a web page. See above.

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: Could be, but doesn't have to be. If someone provides another value to this that will fulfil its name (i.e. it's a blank gif) then there's no reason not to let them!

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: No, because we have to support IE8 with Aikau.

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: Will do

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: Am updating the code currently to support this. Will push the changes shortly.

draperd: It's not at the expense of all reason... a thumbnail (for us) has typically been square, so its not necessary to specify both width and height. As I said, it was shorthand. Width was actually the most important attribute (and the thing easiest controlled). If you like... don't specify width or height and just force the use of dimensions.

MartinDoyleUK:
Have now added support for dimensions object. Also, it's possible to specify an "aspectRatio" config property, so Thumbnail could put in `1` here, and it would automatically mirror any missing width/height property from the other one.

aikau/src/main/resources/alfresco/html/Image.js
draperd: My point is about why have a default a separate attribute, you have two attributes for the same dimension... why not just have one with a default. Would you expect the user to change the default minimum size but not the actual size? That seems odd.

MartinDoyleUK: This is maybe a poorly named variable then. It's only used when no image (or image class) has been provided, to ensure that something is visible on the page. The default size of an image is its natural size.

aikau/src/main/resources/alfresco/html/Image.js
MartinDoyleUK: See new comment at bottom of PR

aikau/src/main/resources/alfresco/html/Image.js
draperd: What is the end result for IE8 if using the approach used in Thumbnail (which is only used when cropToFit is applied). Is it broken or ugly? If ugly (as said previously) we can live with.

MartinDoyleUK: Broken, I think, in the sense that background images will not display because the natural width/height is needed to make the blank.gif image big enough to show them.

draperd: We're providing the dual capability of explicit image source and CSS background image aren't we? (as with Logo currently - I think we agreed on this)... in which case I'm comfortable with this being a limited feature not supported for IE8. If we use the image source approach for carousel buttons then this won't be an issue as it would work in IE8 for Share.

MartinDoyleUK: Sorry. Have looked at the code again. If you don't do this, then background images in all browser will not be resized appropriately, also it will __only__ do this for background images or IE8. I think if someone wants to put an image in, then it should automatically size.

MartinDoyleUK: This won't do any extra network requests, because the image will already be cached. The memory will be almost instantly be marked for GC. It will only do it once for an image. I really don't see this as a problem. To be honest, I see it as a real feature that you can specify a background image and let the code handle the sizing for you!

PR #627, opened 2015-10-23
Removed transition from widget debug

kevinroast:
Removed transition from widget debug - as is causing some UI elements to transition on initial display when client-debug is true - which is distracting and a waste of battery power.

### DIFF
--- a/aikau/src/main/resources/alfresco/debug/css/WidgetInfo.css
+++ b/aikau/src/main/resources/alfresco/debug/css/WidgetInfo.css
@@ -20,7 +20,6 @@
 
 .alfresco-debug-Info {
    &.highlight {
-      transition: padding .5s cubic-bezier(.25, .75, .5, .9);
    }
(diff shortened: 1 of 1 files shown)


draperd:
I agree that it shouldn't affect any other widgets when in debug mode... however, is it not useful to see the explosion/reconstruction when toggling the Developer View? What I was to make the transition specific enough to not impact anything else?

kevinroast:
The developer mode is an edge case. I would not agree that the transition is worth it if it affects elements when developer mode is off. If the transition can be changed to not cause rendering effects when developer mode is off then perhaps that is ok. Otherwise I would remove it.

The other possibility is to make developer mode a page refresh style toggle like SurfBug i.e. it does not decorate the page until refresh, this would certainly mean the CSS wouldn't affect anything at the initial page render, and avoid adding a large number of extra DIVs to the page that are 99% of the time never seen or used.

PR #633, opened 2015-10-27
AKU-658: Update CrudService to support scoped responses

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-658 to update the alfresco/services/CrudService to ensure that requests published from scoped originators use the responseScope when making refresh requests. The unit test has been updated to take advantage of the DebugLog and new tests have been added for scoped originators of publications.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/CrudService.js
+++ b/aikau/src/main/resources/alfresco/services/CrudService.js
@@ -89,12 +89,15 @@ define(["dojo/_base/declare",
          });
 
          var noRefresh = lang.getObject("data.noRefresh", false, originalRequestConfig);
-         if (noRefresh === true) {
+         if (noRefresh === true) 
+         {
             // Don't make a refresh request...
-         } else {
+         } 
+         else 
+         {
             // When refreshing, check to see if a pubSubScope was provided...
-            var pubSubScope = lang.getObject("data.pubSubScope", false, originalRequestConfig) || "";
-            this.alfPublish(pubSubScope + "ALF_DOCLIST_RELOAD_DATA");
+            var pubSubScope = lang.getObject("responseScope", false, originalRequestConfig) || "";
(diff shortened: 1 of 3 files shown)


aikau/src/main/resources/alfresco/services/CrudService.js
```diff
@@ -89,12 +89,15 @@ define(["dojo/_base/declare",
...
-         } else {
+         } 
+         else 
+         {
             // When refreshing, check to see if a pubSubScope was provided...
-            var pubSubScope = lang.getObject("data.pubSubScope", false, originalRequestConfig) || "";
-            this.alfPublish(pubSubScope + "ALF_DOCLIST_RELOAD_DATA");
+            var pubSubScope = lang.getObject("responseScope", false, originalRequestConfig) || "";
+            this.alfPublish("ALF_DOCLIST_RELOAD_DATA", null, false, false, pubSubScope);
          }
```
MartinDoyleUK: This all seems very verbose. The entire block could be rewritten as ...

```javascript
if (noRefresh !== true)
{
  this.alfPublish("ALF_DOCLIST_RELOAD_DATA", null, false, false, originalRequestConfig.alfResponseScope);
}
```

Of additional note, it would be good to use `topics.RELOAD_DATA_TOPIC`, and also shouldn't `alfResponseScope` be used instead of `responseScope`, as we should be taking the final response scope property from the payload, not the one passed into it?

draperd: I've changed the code, but we have to use responseScope and not alfResponseScope because alfResponseScope should only be set by the Core... the general practice agreed was that when publishing we should only set responseScope and not alfResponseScope... as it was, I tried initially with alfResponseScope (from copying your suggestion exactly) and the unit test failed.

MartinDoyleUK: I wasn't sure where the object was coming from, so obviously whatever works!

PR #634, opened 2015-10-29
AKU-655: Ensure delete (and other actions use correct response scope)

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-655. The initially reported bug was that delete was not working correctly, however I also uncovered discrepancies between actions for raw data (directly from the Repository) vs data annotated by the Share tier (to include actions). There was also differences between views that generate pubSubScopes on rows (i.e. the detailed view) and those that don't. This PR has hopefully ironed out all these issues.


aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
```diff
@@ -219,12 +219,17 @@ define(["dojo/_base/declare",
          array.forEach(this.widgetsForActions, function(action) {
             if (action && action.name)
             {
-
+               // It's expected that an Actions rendererer will be used inside a row of Document List view. The
+               // row itself may or may note have its own scope and where some actions will require that the
+               // list is refreshed (e.g. when items are deleted, moved or copied) in which case the responseScope
+               // should be addressed at the list. We therefore need to set the response scope accordingly...
+               var responseScope = this.publishToParent ? this.parentPubSubScope : this.pubSubScope;
```
MartinDoyleUK: I'm sure it's already been considered, but just to double-check, the response scope definitely won't be correctly generated automatically when the alfPublish is called?

draperd: That's not the point.... the responseScope needs to be set according to the requirements of the actions widget. It will take the scope, but not the correct one - we are *choosing* the appropriate response scope according to the configuration.

MartinDoyleUK: That's fine. As I said, I assumed it had been considered already.

PR #638, opened 2015-10-30
AKU-653: Upload to row

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-653 to add support for mixing the alfresco/documentlibrary/_AlfDndDocumentUploadMixin into the alfresco/lists/views/layouts/Row. Unfortunately this required some reworking of the mixin because previously we were creating the drag overlay (to display the highlighting) in the mixing widgets parent node. This didn't work for Row because there were multiple overlays at the same node and they were positioned incorrectly and the layout was corrupted. So the overlays are now created in the body. In order to get this working we've needed to use pointer-events none, but as this is a feature only required for releases that support IE10 an up this is acceptable.

### DIFF
--- a/aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/lists/DNDUploadBehaviour.get.js
+++ b/aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/lists/DNDUploadBehaviour.get.js
@@ -1,3 +1,4 @@
+// jshint maxlen:false
 // Data object to use for the lists/views...
 var data = {
    items: [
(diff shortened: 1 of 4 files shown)


aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/lists/DNDUploadBehaviour.get.js
```diff
@@ -24,15 +45,19 @@ var data = {
...
-function getThumbnail(id, isDraggable) {
+function getThumbnail(options) {
    var thumbnail = {
-      id: id + "_THUMBNAIL",
+      id: options.id + "_THUMBNAIL",
       name: "alfresco/renderers/Thumbnail",
       config: {}
    };
-   if (isDraggable === true || isDraggable === false)
+   if (options.dragableThumbnail === true || options.dragableThumbnail === false)
```
MartinDoyleUK: dragable?

PR #640, opened 2015-10-30
AKU-654: Added alfresco/layout/UploadContainer

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-654 to create a new alfresco/layout/UploadContainer widget that can be used for capturing drag-and-drop uploads anywhere on a page (if all widgets are placed within it)


aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
```diff
@@ -524,19 +524,31 @@ define(["dojo/_base/declare",
...
+       * This sets the position and dimensions of the 
+       * [dragAndDropOverlayNode]{@link module:alfresco/documentlibrary/_AlfDndDocumentUploadMixin#dragAndDropOverlayNode}
+       * 
+       * @instance
+       * @overridable
+       * @since 1.0.42
+       */
+      setDndHighlightDimensions: function alfresco_documentlibrary__AlfDndDocumentUploadMixin__setDndHighlightDimensions() {
+         var computedStyle = domStyle.getComputedStyle(this.dragAndDropNode);
+         var dndNodeDimensions = domGeom.getMarginBox(this.dragAndDropNode, computedStyle);
```
MartinDoyleUK: If this function isn't passed a computed style, then it will retrieve one, exactly as on the line above, so there's no benefit in this instance to doing it here first. It would definitely be of benefit if it were used more than once. As it is, there's no downside to doing it separately, so this is just for future reference.

PR #651, opened 2015-11-03
AKU-663: Path tree node highlight on expand

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-663 to ensure that the PathTree highlights nodes when they are expanded. The unit has been updated and the PathTree has also been updated to improve the way in which tree nodes are given IDs to assist with unit testing

### DIFF
--- a/aikau/src/main/resources/alfresco/navigation/PathTree.js
+++ b/aikau/src/main/resources/alfresco/navigation/PathTree.js
@@ -130,6 +130,11 @@ define(["dojo/_base/declare",
                this.tree._expandNode(targetChildNode);
                this.expandPathElement(targetChildNode, pathElements);
             }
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/navigation/PathTree.js
```diff
@@ -130,6 +130,12 @@ define(["dojo/_base/declare",
                this.tree._expandNode(targetChildNode);
                this.expandPathElement(targetChildNode, pathElements);
             }
+            else
+            {
+               // Focus the last expanded node
+               // this.tree.focusNode(node);
```
MartinDoyleUK: Remove the commented-out line?

PR #656, opened 2015-11-04
AKU-668: Vary-label single button on upload dialog

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-668 to ensure that there is only a single button on the upload dialog. Whilst uploads are in progress this is labelled "Cancel" and once all uploads have completed (regardless of result) it is changed to say "OK" (to indicate that cancellation is no longer possible).

### DIFF
--- a/aikau/src/main/resources/alfresco/upload/AlfUploadDisplay.js
+++ b/aikau/src/main/resources/alfresco/upload/AlfUploadDisplay.js
@@ -18,9 +18,6 @@
  */
 
 /**
- * This module is currently in BETA. Outstanding work:
- * - overall styling design and CSS
- *
  * This module provides simple display handling for file uploads. It displays completed, in progress and failed
  * file uploads in separate sections along with a displayed value of the overall upload progress. 
  *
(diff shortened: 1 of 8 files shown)


aikau/src/main/resources/alfresco/upload/AlfUploadDisplay.js
```diff
@@ -216,12 +213,12 @@ define(["dojo/_base/declare",
...
          }, this.failedItemsNode);
-         var fileName = domConstruct.create("td", {
+         domConstruct.create("td", {
              innerHTML: fileName,
              className: "filename"
          }, failedFile);
-         var reason = domConstruct.create("td", {
-             innerHTML: (error != null && error.reason != null) ? error.reason : message.get("unknown.failure.reason"),
+         domConstruct.create("td", {
+             innerHTML: (error && error.reason) ? error.reason : this.message("unknown.failure.reason"),
```
MartinDoyleUK: Could we need to `this.message` the `error.reason`?? As usual, presume this is correct, but just double-checking.

draperd: It's because it's already been through this.message (see the "handleFailedUpload" that calls this one)

PR #657, opened 2015-11-05
AKU-681: RenderFilter updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-681 to provide the ability to perform token substitution in renderFilter configuration.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/ObjectProcessingMixin.js
+++ b/aikau/src/main/resources/alfresco/core/ObjectProcessingMixin.js
@@ -111,6 +111,19 @@ define(["dojo/_base/declare",
       },
 
       /**
+       * Wrapper for processTokens, searching within currentMetadata
+       *
+       * @instance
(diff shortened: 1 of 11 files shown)


aikau/src/main/resources/alfresco/core/ObjectProcessingMixin.js
```diff
@@ -111,6 +111,19 @@ define(["dojo/_base/declare",
       },
 
       /**
+       * Wrapper for processTokens, searching within currentMetadata
+       *
+       * @instance
+       * @param {string} v The value to process.
```
MartinDoyleUK: Not a big fan of single-letter variables (apart from in for loops) ... makes it immediately slower to read than a more explicit name such as `valueToProcess`, which makes it instantly understandable without having to look at the documentation and/or the way the variable is used, in order to determine what it is.

PR #658, opened 2015-11-05
AKU-677: Smart download action

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-677 to provide a "smart" download for multiple selection actions such that a single selected document is downloaded as the document, but multiple selected documents are downloaded as an archive.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -322,21 +322,6 @@ define([],function() {
       DOWNLOAD_NODE: "ALF_DOWNLOAD_FILE",
 
       /**
-       * This is the response topic for requests to obtain node data that will be immediately
-       * followed by the downloading of that node. It is set as the response topic for smart
-       * downloads of nodes included in Search API requests where the "nodeRef" attribute
-       * is available but the "contentURL" attribute is not. This is not intended to be used by 
-       * widgets or services other than the 
-       * [DocumentService]{@link module:alfresco/services/DocumentService}.
-       *
-       * @instance
-       * @type {string}
-       * @default
-       * @since 1.0.43
-       */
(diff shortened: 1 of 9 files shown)


aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -322,6 +322,36 @@ define([],function() {
...
+       * downloads of nodes included in Search API requests where the "nodeRef" attribute
+       * is available but the "contentURL" attribute is not. This is not intended to be used by 
+       * widgets or services other than the 
+       * [DocumentService]{@link module:alfresco/services/DocumentService}.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.43
+       */
```
MartinDoyleUK: missing @event

aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -322,6 +322,36 @@ define([],function() {
...
+       * [DocumentService]{@link module:alfresco/services/DocumentService}.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.43
+       *
+       * @event
+       */
+      DOWNLOAD_ON_NODE_RETRIEVAL_SUCCESS: "ALF_DOWNLOAD_ON_NODE_RETRIEVAL_SUCCESS",
```
MartinDoyleUK: The handler function needs `response.item`, but this isn't mentioned in property tags. I'm not sure it's a good idea to let any required payload elements be implicit (i.e. not explicitly mentioned).

aikau/src/main/resources/alfresco/core/topics.js
draperd: I'm not sure this is an event... since it is neither subscribed to, nor published... it is simply a response scope. The actual subscription and publication is the DOWNLOAD_ON_NODE_RETRIEVAL_SUCCESS which is why I didn't give it the @event tag.

PR #659, opened 2015-11-05
AKU-680: Faceted search filter encoding updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-680 to implement the recommended changes to support encoding of facet filters described by https://issues.alfresco.com/jira/browse/MNT-14333. One unit test needed to be updated to handle the encoding. I've also verified that the changes work fine with Share 5.0.x releases for backwards compatibility.

### DIFF
--- a/aikau/src/main/resources/alfresco/search/AlfSearchList.js
+++ b/aikau/src/main/resources/alfresco/search/AlfSearchList.js
@@ -468,7 +468,8 @@ define(["dojo/_base/declare",
        */
       updateFilterHash: function alfresco_search_AlfSearchList__updateFilterHash(fullFilter, mode) {
          // Get the existing hash and extract the individual facetFilters into an array
-         var aHash = hashUtils.getHash(),
+         var aHashString = hashUtils.getHashString(),
+             aHash = ioQuery.queryToObject(aHashString),
              facetFilters = aHash.facetFilters ? aHash.facetFilters : "",
              facetFiltersArr = facetFilters === "" ? [] : facetFilters.split(",");
 
@@ -705,7 +706,7 @@ define(["dojo/_base/declare",
 
          // TODO: This should probably be in the SearchService... but will leave here for now...
          var facets = lang.getObject("response.facets", false, payload);
-         var filters = lang.getObject("requestConfig.query.filters", false, payload);
+         var filters = lang.getObject("requestConfig.query.encodedFilters", false, payload);
          if (facets !== null && facets !== undefined)
          {
             for (var key in facets)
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/search/AlfSearchList.js
```diff
@@ -468,7 +468,8 @@ define(["dojo/_base/declare",
        */
       updateFilterHash: function alfresco_search_AlfSearchList__updateFilterHash(fullFilter, mode) {
          // Get the existing hash and extract the individual facetFilters into an array
-         var aHash = hashUtils.getHash(),
+         var aHashString = hashUtils.getHashString(),
+             aHash = ioQuery.queryToObject(aHashString),
```
MartinDoyleUK: It feels a bit wrong to say "this utility isn't doing what I want, so I'll get the bits I need and do it outside the utility instead". Maybe we could just modify the utility getHash and setHash functions to take optional suppressDecoding/suppresssEncoding parameters (respectively), meaning the functionality could be re-used if necessary?

aikau/src/main/resources/alfresco/search/FacetFilter.js
```diff
@@ -253,7 +253,8 @@ define(["dojo/_base/declare",
        */
       _updateHash: function alfresco_search_FacetFilter___updateHash(fullFilter, mode) {
          // Get the existing hash and extract the individual facetFilters into an array
-         var aHash = hashUtils.getHash(),
+         var aHashString = hashUtils.getHashString(),
+             aHash = ioQuery.queryToObject(aHashString),
```
MartinDoyleUK: As above (and already duplication of the functionality, which is a reason to modify it in the utility)

aikau/src/main/resources/alfresco/search/AlfSearchList.js
draperd: No... we can't do this... read through the associated MNT issue... we need to support two versions of Share

aikau/src/main/resources/alfresco/util/hashUtils.js
```diff
@@ -34,12 +34,15 @@ define(["dojo/_base/array",
...
 
       // See API below
-      getHash: function alfresco_util_hashUtils__getHash() {
+      getHash: function alfresco_util_hashUtils__getHash(suppressDecoding) {
          var hashString = this.getHashString(),
             hashObj = ioQuery.queryToObject(hashString);
-         array.forEach(Object.keys(hashObj), function(hashKey) {
-            hashObj[hashKey] = decodeURIComponent(hashObj[hashKey]);
-         });
+         if (suppressDecoding !== true)
```
MartinDoyleUK: Really don't like this. It's totally subverting JavaScript's ability to use truthy/falsy values, and it's counter-intuitive. If I wanted to call something like `getHash(myOptions.length)` then it would only work if I prefixed my passed-through value with `!!`. I know it's used a lot elsewhere, so if you want to keep it then I will merge it as-is.

PR #660, opened 2015-11-06
AKU-671: Dialogs not re-centered on content change

MartinDoyleUK:
This PR addresses [AKU-671](https://issues.alfresco.com/jira/browse/AKU-671) "Dialogs not re-centered on content change" by updating and then applying the updated `ResizeMixin.addResizeListener()` functionality to [`AlfDialog`](https://github.com/Alfresco/Aikau/blob/develop/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js).

### DIFF
--- a/aikau/src/main/resources/alfresco/core/ResizeMixin.js
+++ b/aikau/src/main/resources/alfresco/core/ResizeMixin.js
@@ -110,7 +110,8 @@ define(["dojo/_base/declare",
 
       /**
        * Setup a listener that will call [alfPublishResizeEvent]{@link module:alfresco/core/ResizeMixin#alfPublishResizeEvent}
-       * whenever a resize is detected in the specified node.
+       * whenever a resize is detected in the specified node. If there is an own method on the instance, then it will be called,
+       * so that the listener is removed on widget destruction (encapsulated here to avoid widespread boilerplate-code duplication).
        *
        * @instance
        * @param {object} monitoredNode The node to monitor
@@ -119,26 +120,10 @@ define(["dojo/_base/declare",
        * @since 1.0.42
        */
       addResizeListener: function alfresco_core_ResizeMixin__addResizeListener(monitoredNode, resizeNode) {
-         var resizeHandler = lang.hitch(this, this.alfPublishResizeEvent, resizeNode || monitoredNode.parentNode);
-         return domUtils.addPollingResizeListener(monitoredNode, resizeHandler);
-      },
-
-      /**
-       * A node has been resized, but we don't know whether we need to publish an event. This
-       * function will check for an ancestor "resizable", and publish an event if found, with that
-       * resizable's parent as the resizing node.
-       *
-       * @instance
-       * @param {object} node The node to be checked
-       */
-      onNodeResized: function alfresco_core_ResizeMixin__onNodeResized(node) {
-         var parent = node;
-         while ((parent = parent.parentNode)) {
-            if (domClass.contains(parent, "alfresco-core-ResizeMixin--resizable")) {
-               this.alfPublishResizeEvent(parent.parentNode);
-               break;
-            }
-         }
+         var resizeHandler = lang.hitch(this, this.alfPublishResizeEvent, resizeNode || monitoredNode.parentNode),
+            resizeListener = domUtils.addPollingResizeListener(monitoredNode, resizeHandler);
+         this.own && this.own(resizeListener);
+         return resizeListener;
       }
    });
 });
--- a/aikau/src/main/resources/alfresco/core/ResizeMixin.js
+++ b/aikau/src/main/resources/alfresco/core/ResizeMixin.js
@@ -34,11 +34,13 @@ define(["dojo/_base/declare",
         "alfresco/core/_EventsMixin",
         "alfresco/core/topics",
         "alfresco/util/domUtils",
+        "alfresco/util/functionUtils",
         "dojo/_base/lang",
         "dojo/on",
-        "dojo/dom"], 
-        function( declare, _EventsMixin, topics, domUtils, lang, on, dom) {
-   
+        "dojo/dom",
+        "dojo/dom-class"], 
+        function(declare, _EventsMixin, topics, domUtils, funcUtils, lang, on, dom, domClass) {
+
    return declare([_EventsMixin], {
       
       /**
@@ -61,7 +63,9 @@ define(["dojo/_base/declare",
          // Fire a custom event to let contained objects know that the node has been resized.
          this.throttledPublish(this.alfResizeNodeTopic, {
             node: resizedNode
-         }, true);
+         }, true, false, {
+            timeoutMs: 100
+         });
       },
 
       /**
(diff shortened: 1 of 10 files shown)


aikau/src/main/resources/alfresco/core/ResizeMixin.js
```diff
@@ -61,7 +61,9 @@ define(["dojo/_base/declare",
          // Fire a custom event to let contained objects know that the node has been resized.
          this.throttledPublish(this.alfResizeNodeTopic, {
             node: resizedNode
-         }, true);
+         }, true, false, {
+            timeoutMs: 100
```
draperd: It might be nice to make the time out configurable.

aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
```diff
@@ -166,6 +166,16 @@ define(["dojo/_base/declare",
       handleOverflow: true,
 
       /**
+       * A placeholder for the resize-listener that's enabled while the dialog is visible. This
+       * value is set automatically.
+       *
+       * @instance
+       * @readonly
+       * @type {object}
+       */
```
draperd: Since?

aikau/src/main/resources/alfresco/util/hashUtils.js
```diff
@@ -113,6 +113,7 @@ define(["dojo/_base/array",
        * a suppresssDecoding argument of true is provided.
        *
        * @instance
+       * @function
```
draperd: Why are we adding these and what does it do to the JSDoc output?

aikau/src/test/resources/alfresco/services/DialogServiceTest.js
```diff
@@ -46,440 +46,504 @@ define(["intern!object",
...
 
-               .findAllByCssSelector(".alfresco-dialog-AlfDialog")
-                  .then(function(elements) {
-                     assert.lengthOf(elements, 1, "The previous dialog without an ID was not destroyed");
-                  });
-            });
-      },
+                  .findAllByCssSelector(".alfresco-dialog-AlfDialog")
+                     .then(function(elements) {
+                        assert.lengthOf(elements, 2, "The previous dialog (without an ID) was destroyed");
```
draperd: What's caused this to change?

aikau/src/main/resources/alfresco/core/ResizeMixin.js
MartinDoyleUK: Not sure that makes a lot of sense here? This is essentially a black-box event where you're saying "I don't want resize-events to happen any more often than 100ms", and it feels like an arbitrary but intentional number that can safely be chosen by us, and seems especially appropriate as it coincides with the interval where we're checking nodes for size-changes.

aikau/src/main/resources/alfresco/util/hashUtils.js
MartinDoyleUK: Because they're references to functions and not actually functions, JSDoc needs this tag to know that they are indeed functions. The other utils classes have these already, where appropriate, and they were just missing on this class.

aikau/src/test/resources/alfresco/services/DialogServiceTest.js
MartinDoyleUK: Your recent commit, I think? If you have a look at the differences with whitespace ignored, then this change disappears, so I think it's just do with re-indenting the file.

draperd: OK... first of all, can you not keep corrupting the indenting as I'm very patiently putting them back each time I find one indented incorrectly. I thought the indenting had been corrupted by that bulk find replace you did - I did realise you were changing them as you were going.

But in this particular case I'd like to know why we were testing for 1 element before but now we're testing for 2.

aikau/src/main/resources/alfresco/util/hashUtils.js
draperd: Have you checked what it does to the JSDoc output. Things might have changed since I first got the JSDoc outputting nicely, but we needed to use certain tags to get the output looking correct - I just want to make sure this doesn't change the output unexpectedly.

MartinDoyleUK: Yes. Seemed ok to me from previewing through "grunt docs".

aikau/src/main/resources/alfresco/core/ResizeMixin.js
draperd: Perhaps, but who are we to decide this. If we abstract to a configurable variable then at least it gives the option. Maybe some people do want resize events to occur more often that 100ms. Maybe there will be cases where some widgets want to check more frequently than than the default and others that don't.

MartinDoyleUK: We're the architects of the framework, who make informed decisions based upon stability/performance considerations. If we do make this configurable, then we'll also need to somehow make the common functionUtils class be configurable, as it's in there that the values of 100, 1000 and 10000 are statically defined as appropriate timeouts for repeating functions, and ResizeMixin relies upon domUtils which relies upon functionUtils to achieve this. It does not seem very viable to do, but if you like we can try and work out a way of this happening.

aikau/src/main/resources/alfresco/util/hashUtils.js
MartinDoyleUK: Pushed. Checked. Seems ok to me.

aikau/src/test/resources/alfresco/services/DialogServiceTest.js
MartinDoyleUK: Yes, it was corrupted by the bulk find/replace. I'm therefore following your pattern of correcting the indenting as I go. That's why it's been reindented.

I haven't changed the test. If you view the differences with whitespace ignored (w=1) as I suggest above, you'll see that it's unchanged from develop.

aikau/src/main/resources/alfresco/core/ResizeMixin.js
draperd: OK, fine.

aikau/src/main/resources/alfresco/util/hashUtils.js
draperd: I don't understand how the referencing works, but I've also checked the JSDoc output and it does seem ok - I just wanted to check.

aikau/src/test/resources/alfresco/services/DialogServiceTest.js
draperd: OK

aikau/src/main/resources/alfresco/util/domUtils.js
```diff
@@ -121,6 +129,67 @@ define([
...
+            nodes[monitorKey] = {
+               domNode: nodeToMonitor,
+               height: nodeToMonitor.offsetHeight,
+               width: nodeToMonitor.offsetWidth,
+               callbackFunc: resizeHandler
+            };
+
+            // If there is no monitoring running, start it up
+            if (!resizeMonitor.removeObj) {
+               resizeMonitor.removeObj = funcUtils.addRepeatingFunction(this._checkNodesForResize, "SHORT");
```
draperd: I'm not especially bothered by this, but I'm surprised you've used string literals and not some kind of enum - this is usually the sort of thing that you pull me up on which I why I've mentioned it.

MartinDoyleUK: I agree. I tried, but couldn't find a good way of making it easily accessible to both the using-code and the calling-code.

MartinDoyleUK: Oh, and it didn't seem important enough to have its own enums class ... unless you think it could??

draperd: No, that's fine then. Just thought I'd mention it.

PR #662, opened 2015-11-06
AKU-673: Dialog should take more space if required

MartinDoyleUK:
This addresses [AKU-673](https://issues.alfresco.com/jira/browse/AKU-673) by allowing passthrough of the fullscreen option when adding a comment. Specifically, one can now specify `addCommentsFullScreen: true` to the config of the `CommentsList` widget.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/CommentsList.js
+++ b/aikau/src/main/resources/alfresco/renderers/CommentsList.js
@@ -49,6 +49,33 @@ define(["dojo/_base/declare",
        * @default [{i18nFile: "./i18n/CommentsList.properties"}]
        */
       i18nRequirements: [{i18nFile: "./i18n/CommentsList.properties"}],
+
+      /**
+       * Whether to use a full-screen dialog when adding comments.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.43
+       */
+      addCommentsFullScreen: false,
+
+      /**
+       * Executed after properties mixed in and before widget created
+       *
+       * @instance
+       * @override
+       * @since 1.0.43
+       */
+      postMixInProperties: function alfresco_renderers_CommentsList__postMixInProperties() {
+         this.inherited(arguments);
+         if (this.addCommentsFullScreen) {
+            var addCommentPayload = lang.getObject("widgets.0.config.widgetsBefore.0.config.publishPayload", false, this);
+            if(addCommentPayload) {
+               addCommentPayload.fullScreenMode = true;
+            }
+         }
+      },
       
       /**
        * Widget has been started, but not necessarily any sub-widgets.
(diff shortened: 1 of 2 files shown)


aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/renderers/CommentsList.get.js
```diff
@@ -19,6 +19,7 @@ model.jsonModel = {
          name: "alfresco/renderers/CommentsList",
          id: "COMMENT_LIST",
          config: {
+            addCommentsFullScreen: true,
```
draperd: I didn't mean that this should be the default... this should be something that the Integrations team can configure when they detect that they're viewing the Document Library in an iframe (so that space is restriced)

aikau/src/main/resources/alfresco/renderers/CommentsList.js
```diff
@@ -49,6 +49,33 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * Whether to use a full-screen dialog when adding comments.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.43
+       */
+      addCommentsFullScreen: false,
```
draperd: We also need to make the full screen dialog padding  ("fullScreenPadding") as well so that the dialog can take up the full space available.

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/renderers/CommentsList.get.js
MartinDoyleUK: It's not the default, this is the test webscript. I will create another CommentsList on the page though, so that both are viewable. Not sure how easily this can be automatically tested? Would we just try and check how much space is around the dialog?

aikau/src/main/resources/alfresco/renderers/CommentsList.js
MartinDoyleUK: So, are you suggesting a second property on CommentsList, or that this `addCommentsFullScreen` property should set both?

draperd: That's an interesting question. For the specific case we're looking at we almost want no padding at all so there's an argument that says we automatically set the padding to be 0... however, for future-proofing it might be worth exposing the additional argument.

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/renderers/CommentsList.get.js
draperd: No, I appreciate that this is the test WebScript - however it would be useful for the test WebScript to represent the default as well as optional configuration parameters (to ensure that both modes work adequately). There is an iframe widget that was added for the analytics work (alfresco/integration/IFrame) that we might be able to make use of here (given that this is the exact use case that we're trying to solve).

MartinDoyleUK:
Need to rebase branch, so will close this PR and create new one when finished.

PR #666, opened 2015-11-09
AKU-690: Focus on widgets in grid without focus function

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-690 to ensure that widgets in an alfresco/lists/views/layouts/Grid can be focused even when they do not have a "focus" function. A unit test has been added to support the fix.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/views/layouts/Grid.js
+++ b/aikau/src/main/resources/alfresco/lists/views/layouts/Grid.js
@@ -42,13 +42,17 @@ define(["dojo/_base/declare",
         "dojo/keys",
         "dojo/_base/lang",
         "dojo/_base/array",
+        "dojo/dom-attr",
+        "dojo/dom-class",
         "dojo/dom-construct",
         "dojo/dom-geometry",
         "dojo/query",
         "dojo/dom-style",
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/lists/views/layouts/Grid.js
```diff
@@ -189,6 +193,35 @@ define(["dojo/_base/declare",
...
+            if(this.focusedChild && widget !== this.focusedChild){
+               this._onChildBlur(this.focusedChild);  // used to be used by _MenuBase
+            }
+            if (widget.domNode)
+            {
+               domAttr.set(widget.domNode, "tabIndex", this.tabIndex); // for IE focus outline to appear, must set tabIndex before focus
+               focusUtil.focus(widget.domNode);
+            }
+         }
+         domClass.add(widget.domNode, "alfresco-lists-views-layouts-Grid--focused");
```
MartinDoyleUK: Are we really changing the state of the grid at this point? The state applies to the element that it follows. Would it be more appropriate to have `alfresco-lists-views-layouts-Grid__cell--focused`?

PR #673, opened 2015-11-12
AKU-670: Auto twister width

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-670 to add a new configuration option for the alfresco/layout/Twister so that it can be configured with a width of "AUTO" to automatically grow and shrink to the available width.


aikau/src/main/resources/alfresco/layout/Twister.js
```diff
@@ -87,11 +87,12 @@ define(["dojo/_base/declare",
         "dojo/_base/array",
         "dojo/dom-construct",
         "dojo/dom-class",
+        "dojo/dom-geometry",
```
MartinDoyleUK: Is this still needed?

PR #681, opened 2015-11-17
AKU-675: Fix expanded cell resize bug

draperd:
This PR fixes a problem observed in the sprint review when demonstrating https://issues.alfresco.com/jira/browse/AKU-675. This ensures that the the expanded panel cell does not get resized when it is displayed. A unit test has been added to verify the fix.


MartinDoyleUK:
Still don't like all this JavaScript sizing, but can't see any problems with the code. Assuming the tests pass, so merging.

PR #685, opened 2015-11-17
AKU-694: Download Cancel is not canceling download

MartinDoyleUK:
This addresses [AKU-694](https://issues.alfresco.com/jira/browse/AKU-694) by changing the cancel-button publish from `ALF_CLOSE_DIALOG` to `ALF_PROGRESS_CANCELLED`. This automatically closes the dialog. Additionally, the cancel handler-function sets a flag to say that the task(s) have been cancelled and for the completion actions not to be run.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/Progress.js
+++ b/aikau/src/main/resources/alfresco/renderers/Progress.js
@@ -68,13 +68,13 @@ define(["dojo/_base/declare",
       templateString: template,
 
       /**
-       * renderProgressUI
+       * The message to display when progress is complete.
        *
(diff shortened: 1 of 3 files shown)


draperd:
Has this been tested against Share at all yet? Or is this just unit test driven?

MartinDoyleUK:
Just unit-tests

draperd:
Code looks good though... I'm happy to merge, but would be good to get the sniff test against Share. I can do that if you like.

MartinDoyleUK:
Thank you ... would be good if you could! :)

PR #690, opened 2015-11-20
AKU-696: Scrolling in dialog scrolls in window

MartinDoyleUK:
This addresses [AKU-696](https://issues.alfresco.com/jira/browse/AKU-696) by setting overflow:hidden on a 100% height top-container when dialogs are visible. Also removed some older attempts to fix this. No unit-tests applicable for this change, but have tested manually using the DialogService test page in Chrome, FF and IE9.

### DIFF
--- a/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
+++ b/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
@@ -408,7 +408,6 @@ define(["dojo/_base/declare",
        */
       onHide: function alfresco_dialogs_AlfDialog__onHide() {
          this.inherited(arguments);
-         domStyle.set(document.documentElement, "overflow", "");
          domClass.remove(this.domNode, "dialogDisplayed");
          domClass.add(this.domNode, "dialogHidden");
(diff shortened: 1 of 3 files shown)


draperd:
One of the reasons that we wanted the scrollbar to always be visible was because it would cause layout issues as dialogs were shown and hidden. For example with the Document Library the list would occasionally jump *below* the sidebar. Before we merge these changes I want to do some investigation into whether or not this is still an issue - there is no point in fixing one bug if it simply creates another.

MartinDoyleUK:
That sounds like a good idea. Will pick that up Monday morning, definitely.

draperd:
Unfortunately as I suspected this has re-introduced the problem that making the scroll bar always visible was resolving - I've attached a video to the JIRA ticket to demonstrate... this PR can't be merged in its current state.

MartinDoyleUK:
Have pushed a small update to the CSS that appears to fix this. Please would you be able to re-review this and see if it's now suitable for merging?

draperd:
Yes, this change seems to work fine.

PR #696, opened 2015-11-25
AKU-706: Ensure that Actions renderer menu is always visible.

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-706 to ensure that the alfresco/renderers/Actions menu is always visible. In order to resolve this issue it has been necessary to completely change the implementation of the actions menu from using an AlfMenuBar (that would only ever open downwards and therefore could result in menu items disappearing off the screen) to using a dijit/Menu which when clicked will always open so that it can bee seen.

In changing this implementation it was necessary to update many of the unit tests to adjust to the new structure, however the new implementation attempts to match the previous structure as best as possible and in many cases the tests have generally been improved using current best practices.

Finally a unit test has been added to verify that the menu is always visible.


draperd:
Don't merge yet please - still one test failing on full regression run

aikau/src/main/resources/alfresco/renderers/Actions.js
```diff
@@ -131,6 +142,69 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * Handles key press events on the actions renderer and opens the menu if 
+       * 
+       * @instance
+       * @param  {object} evt The key press event
+       * @since 1.0.46
+       */
+      onKeyPress: function alfresco_renderers_Actions__onKeyPress(evt) {
+         if (evt && (evt.charOrCode === keys.ENTER || evt.charCode === keys.SPACE))
```
MartinDoyleUK: * Has `charOrCode` definitely been set? I think you sometimes need to calculate it yourself as a local variable
* The second condition uses `evt.charCode` not `evt.charOrCode`

draperd: During development I observed that charOrCode can reliably be used for ENTER but not for SPACE.... this is why they are using different attributes. I think this is because SPACE has an actual character (i.e. " ") but ENTER doesn't.

PR #701, opened 2015-11-27
Moved calculation of line-height for buttons into vertical padding.

kevinroast:
This avoids the unpleasant looking focus rectangle being rendered around the AlfButton text in a bump and instead renders a clean rectangle shape. Buttons look exactly the same as before when not in focus. No functionality or code changed.

Buttons checked in Share 5.1 to confirm no visual changes other than focus rectangle.

See https://issues.alfresco.com/jira/browse/SFS-275


MartinDoyleUK:
The only possible difference is to do with vertical alignment alongside other elements, however if nothing immediately presents, then I suggest this isn't an immediate concern, rather something to keep an eye on and to probably fix "there" rather than "here" as I agree about the focus rectangle problem!

PR #704, opened 2015-11-30
AKU-731: Double encoding in hashUtils

MartinDoyleUK:
This addresses [AKU-731](https://issues.alfresco.com/jira/browse/AKU-731) by removing the double encoding/decoding present because of the explicit encodeURICompoment and the implict one inside Dojo's ioQuery module, and similarly the decode version. Share checked manually without finding any problems. Full suite run successfully.


draperd:
Reviewed - I'll merge when you let me know that the tests have all passed.

PR #705, opened 2015-11-30
AKU-692: AlfTabContainer in iframe issues

draperd:
This issues addresses https://issues.alfresco.com/jira/browse/AKU-692 to resolve the issue of the AlfTabContainer being loaded within an iframe. It also includes a lot more defensive code around both the AlfTabContainer and the HorizontalWidgets modules to ensure they work appropriately when contained in widgets that are initially invisible. 

A unit test has been created to reproduce the problem and verify that it is solved. This also requires the use of an extension module that has been attached to the JIRA ticket that creates an AMD mapping to a patched version of the dojo/hccss module that was failing in iframes on Firefox.


aikau/src/main/resources/alfresco/layout/AlfTabContainer.js
```diff
@@ -295,80 +295,99 @@ define(["dojo/_base/declare",
...
-               {
-                  // No action, use the explicit configuration
-               }
-               else if (widget.delayProcessing === true && widget.selected)
-               {
-                  // Silly configuration, if the tab is to be initially selected, processing can't be delayed...
-                  widget.delayProcessingDefault = false;
-               }
-               else
+            this.tabContainerWidget = new TabContainer({
```
MartinDoyleUK: `this.create[...]` for the pubSub inheritance? (assuming already considered, but just flagging in case)

draperd: It's encapsulated within the Aikau widget - so it doesn't need any pub/sub data and wouldn't know what to do with it if it was given it (it's a Dojo widget)

aikau/src/main/resources/alfresco/layout/HorizontalWidgets.js
```diff
@@ -157,8 +171,17 @@ define(["alfresco/core/ProcessWidgets",
          // Split the full width between all widgets... 
          // We should update this to allow for specific widget width requests...
          this.visibilityRuleTopics = this.getVisibilityRuleTopics(this.widgets);
-         this.doWidthProcessing(this.widgets, true);
-         this.inherited(arguments);
+
+         // NOTE: Here we're very deliberately using JQuery to get the available width, this
+         //       is to address a very specific issue with Firefox failing when using iframes
+         //       and the dojo/dom-geometry code (in particular the getComputedStyle) - see AKU-692
+         var overallwidth = $(this.domNode).width();
```
MartinDoyleUK: Happy to leave JQuery in if it's tested and working, but curious to know whether `this.domNode.offsetWidth` was tried?

draperd: I didn't, but at this stage I'd rather not change when the current implementation is working fine.

MartinDoyleUK: No, of course not ... would be good to try and use native over library though, if we can, for obvious reasons.

PR #706, opened 2015-12-01
Suggested localization changes from the User Assistance team

draperd:
These are changes to localization messages suggested by the User Assistance team


aikau/src/main/resources/alfresco/services/i18n/CloudSyncService.properties
```diff
@@ -1,22 +1,22 @@
-cloud-auth.dialog.title=Sign in to Alfresco in the cloud
-cloud-auth.dialog.save.label=Your details will be saved for future sync actions
-cloud-auth.dialog.auth.error.title=Authentication Error
+cloud-auth.dialog.title=Sign in to Alfresco in the Cloud
```
MartinDoyleUK: The other uses of `cloud` in this changeset are all lower-case ... is this intentional?

PR #717, opened 2015-12-04
AKU-793: Facet filter config

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-703 to make it possible to make the facet filters have a "full width" click area through configuration.

### DIFF
--- a/aikau/src/main/resources/alfresco/search/FacetFilter.js
+++ b/aikau/src/main/resources/alfresco/search/FacetFilter.js
@@ -124,6 +124,17 @@ define(["dojo/_base/declare",
       filterData: "",
 
       /**
+       * Indicates whether or not the full width of the filter (including the count and the white space)
+       * can be clicked to toggle the filter.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.47
+       */
+      fullWidthClick: false,
+
+      /**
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/search/FacetFilter.js
```diff
@@ -187,14 +203,21 @@ define(["dojo/_base/declare",
        *
        * @instance
        */
-      onToggleFilter: function alfresco_search_FacetFilter__onToggleFilter(/*jshint unused:false*/ evt) {
-         if (this.applied)
+      onToggleFilter: function alfresco_search_FacetFilter__onToggleFilter(evt) {
+         if (evt.target !== this.labelNode && !this.fullWidthClick)
```
MartinDoyleUK: I'm sure this is correct, but just want to be sure this shouldn't be a `||` rather than `&&`?

draperd: Only the label node will support the click event when fullWidthClick is not true... so if fullWidthClick is not enabled then any other node but labelNode will have no impact. The unit tests verify the logic as well.

MartinDoyleUK: This is one of the problems with having negative conditions in an if-else ... more difficult to understand than positive. So, you will change the filters if you've clicked on the label node or full width is turned on. Would prefer to see it that way round.

PR #718, opened 2015-12-04
AKU-717: Add support for TinyMCE editor to auto resize

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-717 to make it possible to the TinyMCE editor to automatically grow to take up available space. In particular this is focused around the full screen commenting support for CommentsList.

### DIFF
--- a/aikau/src/main/resources/alfresco/editors/TinyMCE.js
+++ b/aikau/src/main/resources/alfresco/editors/TinyMCE.js
@@ -37,13 +37,16 @@ define(["dojo/_base/declare",
         "dijit/_TemplatedMixin", 
         "dojo/text!./templates/TinyMCE.html", 
         "alfresco/core/Core", 
+        "alfresco/core/ResizeMixin",
         "alfresco/core/topics",
         "service/constants/Default", 
-        "dojo/_base/lang"], 
-        function(declare, _WidgetBase, _TemplatedMixin, template, AlfCore, topics, AlfConstants, lang) {
+        "dojo/_base/lang",
+        "jquery",
+        "jqueryui"], 
+        function(declare, _WidgetBase, _TemplatedMixin, template, AlfCore, ResizeMixin, topics, AlfConstants, lang, $) {
 
 
-   return declare([_WidgetBase, _TemplatedMixin, AlfCore], {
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/editors/TinyMCE.js
```diff
@@ -209,6 +243,64 @@ define(["dojo/_base/declare",
...
+       * @since 1.0.47
+       */
+      onResize: function alfresco_editors_TinyMCE__onResize() {
+         if (this.autoResize)
+         {
+            if (this._editorInitialized)
+            {
+               // Work our way back up through the current node ancestors to find
+               // the first ancestor that have height and width set on them
+               var height, width;
```
MartinDoyleUK: IndexOf on a style seems a bit strange. Also the code seems a bit off ... it looks like you'll end up with the first height you find and the first width you find, even if not on the same node, which might be intended but isn't what the comment says.

I'd have done this ...

```javascript
var parent = this.domNode,
   size;
while ((parent = parent.parentNode) && !size) {
   if (parent.style.height && parent.style.width) {
      size = {
         h: parent.style.height,
         w: parent.style.width
      };
   }
}
```

Up to you if you want to use any of it, or if you're happy with it as-is, but I'm not sure it's right at the moment.

draperd: What you describe is exactly my intention (e.g. take height and width even if not from the same node). No point in taking a width when there is no height (as that would prevent it growing). The technique I've used is the same I got to work in the calculateParentHeight function from the HeightMixin module so I just used something that I knew worked.

MartinDoyleUK: Ok, then the comments need to be updated, because right now it looks like it's saying get the first ancestor which has height and width set on it.

MartinDoyleUK: Also, it's very bad to rely on the string formatting of style elements returned by the browser (I've never seen that documented anywhere) by doing an indexOf with the style attribute value.

Finally, you ask if there's a width value set on the inline-style of an element and then calculate the height of the element rather than using it, which could potentially give a different value ... is this intentional? If so, then it needs to be documented as it's not logical.

PR #720, opened 2015-12-07
AKU-48: Playground application

draperd:
This PR is the first part of https://issues.alfresco.com/jira/browse/AKU-748. It defines a new application for show-casing and experimenting with Aikau widgets.


aikau-playground-application/src/main/java/org/alfresco/aikau/ExampleList.java
```diff
@@ -0,0 +1,111 @@
...
+   {
+      Path path = getContainer().getRegistry().getFamily("/ExamplePage");
+
+      int ps = 25;
+      String pageSize = req.getParameter("pageSize");
+      if (pageSize != null)
+      {
+         try
+         {
+             ps = Integer.parseInt(pageSize, 10);
```
MartinDoyleUK: Default is already 10 in Java, so don't need to pass second argument.

aikau-playground-application/src/main/java/org/alfresco/aikau/ExampleList.java
```diff
@@ -0,0 +1,111 @@
...
+             // No action required
+         }
+      }
+      int si = 0;
+      String page = req.getParameter("page");
+      if (page != null)
+      {
+         try
+         {
+             int p = Integer.parseInt(page, 10);
```
MartinDoyleUK: As above, the 10 is optional

PR #723, opened 2015-12-09
Enhance 'loading' status of AlfList components to only display message after a configurable delay.

MartinDoyleUK:
This addresses [AKU-700](https://issues.alfresco.com/jira/browse/AKU-700) in several ways. Firstly, views now only get removed when a new one is ready to take its place. Secondly, a loading message is display as an overlay on the views node when new data is loading. Finally, the loading message has had the appropriate variables extracted into defaults.less to be overridden as necessary by implementing projects.


aikau/src/main/resources/alfresco/core/CoreWidgetProcessing.js
```diff
@@ -175,6 +176,7 @@ define(["dojo/_base/declare",
        * @param {array} widgets An array of the widget definitions to instantiate
        * @param {element} rootNode The DOM node which should be used to add instantiated widgets to
        * @param {string} processWidgetsId An optional ID that might have been provided to map the results of the call to
+       * this option should be used with caution.
```
MartinDoyleUK: This seems like a funny line to have?

draperd: Yep, failed to delete that line... I'll remove it

PR #724, opened 2015-12-09
AKU-732 & AKU-745: AlfBreadcrumbTrail updates

MartinDoyleUK:
This addresses [AKU-745](https://issues.alfresco.com/jira/browse/AKU-745) and [AKU-732](https://issues.alfresco.com/jira/browse/AKU-732) by changing NavigationService to modify rather than replace the current hash when a certain flag is passed. The changes also fix the escaping bug. Have also improved hashUtils slightly. Full regression suite run successfully.


draperd:
Don't we need to have some tests that verify this behaviour - i.e. that it is possible to augment a hash?

MartinDoyleUK:
Yes. There should be tests! Will add some.

PR #736, opened 2015-12-11
AKU-754: Dutch specific NLS correction

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-754


draperd:
Thanks Gloria, I've made that additional change

PR #737, opened 2015-12-14
Feature/aikau and dojo cachebust

fkleon:
Using Aikau with [Dojo's cache bust feature](https://dojotoolkit.org/documentation/tutorials/1.10/dojo_config) leads to invalid URLs to some static resources.

I was not able to open an issue on the Alfresco Jira against the Aikau project, hence description and code over here on GitHub.

###### Description

In many Aikau widgets the path to resources is appended to an URL generated by `require.toUrl`. However, when using cache bust Dojo already adds a query parameter to this URL. This leads to broken URLs such as:
`/aikau/res/js/aikau/1.0.48-SNAPSHOT/alfresco/footer?1450127564788/css/images/alfresco-share-logo.png`

I have added an extension JAR which enables cache busting for the test app to demonstrate the problem and fix, but I am not aware of an easy way to selectively load it for certain tests suites only.

###### License

(c) by Catalyst.net Ltd, submitted under the Apache License v2.

### DIFF
--- a/aikau/src/main/resources/alfresco/debug/WidgetInfo.js
+++ b/aikau/src/main/resources/alfresco/debug/WidgetInfo.js
@@ -71,7 +71,7 @@ define(["dojo/_base/declare",
        * @instance
        */
       postMixInProperties: function alfresco_debug_WidgetInfo__postMixInProperties() {
-         this.imgSrc = require.toUrl("alfresco/debug") + "/css/images/info-16.png";
+         this.imgSrc = require.toUrl("alfresco/debug/css/images/info-16.png");
          if (this.displayId)
          {
             this.altText = this.message("widgetInfo.alt.text", {
--- a/aikau/src/main/resources/alfresco/dnd/DroppedItemWrapper.js
+++ b/aikau/src/main/resources/alfresco/dnd/DroppedItemWrapper.js
@@ -185,10 +185,10 @@ define(["dojo/_base/declare",
          this.downAltText = this.encodeHTML(this.message(this.downAltText));
          this.deleteAltText = this.encodeHTML(this.message(this.deleteAltText));
          this.editAltText = this.encodeHTML(this.message(this.editAltText));
-         this.upImageSrc = require.toUrl("alfresco/dnd") + "/css/images/" + this.upImg;
-         this.downImageSrc = require.toUrl("alfresco/dnd") + "/css/images/" + this.downImg;
-         this.editImageSrc = require.toUrl("alfresco/dnd") + "/css/images/" + this.editImg;
-         this.deleteImageSrc = require.toUrl("alfresco/dnd") + "/css/images/" + this.deleteImg;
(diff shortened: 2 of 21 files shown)


draperd:
Thanks for submitting this pull request.. We'll need to review the changes in full and run all the regression tests before merging although it looks like a sensible change to make. 

I also need to review the fact that you've added a license statement to the PR... I've not seen that before and need to double check that this doesn't cause us any problems going forwards.

draperd:
Also, I understand how this could be useful for image downloads, however I also want to double check on the implications on module loading since the Surf framework that Aikau relies upon is already performing cache busting by aggregating modules into a single resource that uses an MD5 checksum of the contents as its contents (this means that if any of the modules in the aggregated resource change that a different resource will need to be requested). So this change will address image loading, but have no impact on modules - therefore I want to be sure that this doesn't degrade performance in any way - particulate if this cache busting will *break* the fact that we can allow our generated resources to be cached indefinitely.

fkleon:
Hi Dave, thanks for your reply.

The JAR functions very similar to the `AMDModuleMapping.jar` and contains just a module definition and a dojo-bootstrap file initialising the cacheBust parameter in dojoConfig with a simple `Date.now()` - on a production environment this would be something like the build ID. I have included it mainly to demonstrate and replicate the problem (by turning on cache busting), and would be happy to remove it again if it turns out to be hard to maintain or is out of place here. However, including it does not interfere with cache busting for any given Aikau application, but only for the testApp bundled here - and possibly the performance of the test suites.

Yes - Surf is already doing a pretty good job at cache busting with it's MD5 hashes. It however seems to exclude some resources such as static files coming from Dojo/Aikau. In production, those files are very unlikely to change anyway. We ran into this problem only when we enabled Dojo's cache busting for development purposes, and it actually broke some of Share's UI starting with Alfresco 5.0.d. Given that nobody else reported this, it is probably a somewhat rare setup.

We have chosen to include the Apache license as per [Alfresco's Contribution Guidelines](https://wiki.alfresco.com/wiki/Contribution_Guidelines), it should be permissive enough to avoid signing a special contribution agreement.

draperd:
I'm sorry we've not got to this yet, and probably won't until the new year now... I'd happily pull the request *without* the JAR file - but I'm concerned about adding the cache busting to the unit test application. That's my main concern at the moment. If you just want to get the main code changes in so that you can take advantage of the cache busting yourself then please amend the PR, otherwise we'll get to this early in the New Year (we weren't working on Aikau last week and we're about to break for Xmas so there won't be much happening over the next couple of weeks).

fkleon:
I've removed the JAR from this PR for now - feel free to merge it.

If you are interested in looking at the cache bust again in the new year, the contents of the JAR are [in this gist](https://gist.github.com/fkleon/ca12c8f370bb5cda7e02) and the original commit is in [this feature branch](https://github.com/fkleon/Aikau/tree/feature/aikau-and-dojo-cachebust-with-jar) on my fork.

PR #738, opened 2015-12-21
AKU-747: Create non-modal fixed position dialog

MartinDoyleUK:
This addresses [AKU-747](https://issues.alfresco.com/jira/browse/AKU-747) and creates a fixed-position StickyPanel widget. Tests have been created accordingly.

Full suite has been run and only `src/test/resources/alfresco/dnd/NestedConfigurationTest` has any failures, which seems entirely unrelated to any of this work, so should be safe to merge. Would be good to have a look at the failures though, to see if there's a quick fix?

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -272,8 +272,29 @@ define([],function() {
        * @instance
        * @type {string}
        * @default
+       *
+       * @event
+       * @property {string} message The message to be displayed in the prompt
+       * @property {string} [title] The title of the prompt
        */
       DISPLAY_PROMPT: "ALF_DISPLAY_PROMPT",
+
+      /**
+       * This topic can be published to request that a [StickyPanel]{@link module:alfresco/layout/StickyPanel} be
+       * displayed. It is subscribed to by the [NotificationService]{@link module:alfresco/services/NotificationService}.
+       *
+       * @instance
+       * @type {string}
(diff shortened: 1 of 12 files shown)


aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -808,6 +829,43 @@ define([],function() {
...
+       * This can be called to close the StickyPanel.
+       *
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.48
+       *
+       * @event
+       */
+      STICKY_PANEL_CLOSE: "STICKY_PANEL_CLOSE",
```
draperd: Can we prefix all of the string values with "ALF" please - less likely for any 3rd party string clashes that way

aikau/src/main/resources/alfresco/layout/css/StickyPanel.css
```diff
@@ -0,0 +1,68 @@
+.alfresco-layout-StickyPanel {
+   bottom: 0;
+   left: 50%;
+   position: fixed;
+   right: 0;
+   &__panel {
+      background: #fff;
```
draperd: LESS variable? I think we have a primary background colour one that could be used as a seed for a stick panel one. I'm not hugely fussed about this for this first cut as I know we're going to be iterating on it.

I'll probably raise another feature request to LESS-ify the whole thing - so no need to make changes, just wanted to note it.

PR #744, opened 2015-12-23
AKU-751: Focus on first field in form

MartinDoyleUK:
This addresses [AKU-751](https://issues.alfresco.com/jira/browse/AKU-751) and will focus on the first field in the form, on page load, if the following is added to the form config:

```javascript
firstFieldFocusOnLoad: true,
```

Unit-test added.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/Form.js
+++ b/aikau/src/main/resources/alfresco/forms/Form.js
@@ -1053,6 +1053,8 @@ define(["dojo/_base/declare",
 
             // If configured, focus the first field in the form
             if (this.firstFieldFocusOnLoad) {
+               // Use setTimeout to allow other synchronous processes to complete first, as this
+               // has a much greater chance of successfully focusing on the field
                setTimeout(lang.hitch(this, this.focus));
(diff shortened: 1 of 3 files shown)


aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -1037,6 +1050,11 @@ define(["dojo/_base/declare",
                this.publishValidValue();
             }
             this._formSetupComplete = true;
+
+            // If configured, focus the first field in the form
+            if (this.firstFieldFocusOnLoad) {
+               setTimeout(lang.hitch(this, this.focus));
```
draperd: Why the time out here?

MartinDoyleUK: Didn't work without it. Worked reliably with it. Figure other processes are running on the page. This lets every synchronous process run and then executes the focus.

draperd: It might be worth adding a comment to that effect for future reference.

PR #750, opened 2016-01-05
AKU-770: Ensure PathTree content creation reloads work for root nodes

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-770 to ensure that root tree nodes in the PathTree can be refreshed on content creation topic publications. The unit tests have been updated to verify this works - delete was already working but a test has been added to be sure and prevent regressions.

### DIFF
--- a/aikau/src/test/resources/alfresco/navigation/PathTreeTest.js
+++ b/aikau/src/test/resources/alfresco/navigation/PathTreeTest.js
@@ -170,7 +170,7 @@ define(["intern!object",
                .getLastXhr()
                   .then(function(xhr) {
                      /* jshint maxlen:500*/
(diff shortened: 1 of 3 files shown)


aikau/src/test/resources/alfresco/navigation/PathTreeTest.js
```diff
@@ -187,6 +187,33 @@ define(["intern!object",
...
+         "Check root node refresh on folder delete": function() {
+            return browser.findById("DELETE_FOLDER_AT_ROOT_label")
+                  .clearXhrLog()
+                  .click()
+               .end()
+
+               .getLastXhr()
+                  .then(function(xhr) {
+                     /* jshint maxlen:500*/
+                     assert.include(xhr.request.url, "/slingshot/doclib/treenode/node/alfresco/company/home/?perms=false&children=false&max=500&libraryRoot=workspace%3A%2F%2FSpacesStore%2Fb4cff62a-664d-4d45-9302-98723eac1319", "Node refresh not requested");
```
MartinDoyleUK: Not sure the error message is necessary here. There's a single assert and the name of the test should be enough to know what's failing.

PR #751, opened 2016-01-05
AKU-764: Base form test updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-764 to tidy up the base form control test and ensure that pubSubScopes are set correctly. I've added a warning if a control within a form is given an explicit pubSubScope - I've resisted the urge to remove it completely.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/Form.js
+++ b/aikau/src/main/resources/alfresco/forms/Form.js
@@ -401,7 +401,7 @@ define(["dojo/_base/declare",
                   widget.config.showValidationErrorsImmediately = this.showValidationErrorsImmediately;
                   if (widget.config.pubSubScope)
                   {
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/forms/Form.js
```diff
@@ -399,6 +399,10 @@ define(["dojo/_base/declare",
                if (widget && widget.config)
                {
                   widget.config.showValidationErrorsImmediately = this.showValidationErrorsImmediately;
+                  if (widget.config.pubSubScope)
+                  {
+                     this.alfLog("warn", "It is not recommended to set a pubSubScope attribute on a form control nested within a form - the pubSubScope should be inherited!", this, widget);
```
MartinDoyleUK: Why is it not recommended? No mention that the field won't be submitted anywhere?

draperd: Good point

PR #754, opened 2016-01-07
AKU-748: Sandpit application updates.

draperd:
This PR relates to https://issues.alfresco.com/jira/browse/AKU-748 to make some updates to the Aikau Sandpit. The main change is to support the capability for mock XHR data (so that we can have data available for lists, etc). This has required adding the Sinon lib files into the application and moving some of the testing widgets into the main Aikau code base. A basic list example has been added to verify that the mock capabilities work.


aikau/src/main/resources/alfresco/testing/MockXhr.js
```diff
@@ -73,14 +79,19 @@ define(["dojo/_base/declare",
...
          if (this.respondAfter)
          {
             this.server.autoRespondAfter = this.respondAfter;
          }
+         server.autoRespond = true;
+         server.xhr.useFilters = true;
+
+         // Adds a filter to allow dynamic dependency requests to be processed normally...
+         server.xhr.addFilter(function(method, url) {
+           return !!url.match(/surf\/dojo\/xhr\/dependencies/);
```
MartinDoyleUK: Purely for academic reasons, you might be interested to know that it'd be twice as fast to use RegExp.exec ...

```javascript
return !!/surf\/dojo\/xhr\/dependencies/.exec(url);
```

draperd: Good to know. No change required though? Speed isn't an issue here.

PR #766, opened 2016-01-12
AKU-775: Ensure TinyMCE resize works effectively

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-775. Because of the lack of proper AMD support for TinyMCE we're forced to load the resources on demand. This means that the first time we create a TinyMCE editor we go through a slightly different code path and for some reason the timing of this results in a dimension calculation error. The simplest resolution to this problem is to add a short timeout on this initial code path that results in the TinyMCE editor being resized correctly. 

This PR also updates the related CommentsList test to make it more reliable on Chrome.

### DIFF
--- a/aikau/src/main/resources/alfresco/editors/TinyMCE.js
+++ b/aikau/src/main/resources/alfresco/editors/TinyMCE.js
@@ -391,7 +391,8 @@ define(["dojo/_base/declare",
          if (this._resizeWhenInitialized)
          {
             // See AKU-775 - adding a short timeout here allows everything to complete initialization before resizing
-            setTimeout(lang.hitch(this, this.onResize), 100);
+            //               by using timeout without a time we can delay until all synchronous processes have completed.
+            setTimeout(lang.hitch(this, this.onResize));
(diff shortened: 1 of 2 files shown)


aikau/src/main/resources/alfresco/editors/TinyMCE.js
```diff
@@ -390,7 +390,8 @@ define(["dojo/_base/declare",
          }
          if (this._resizeWhenInitialized)
          {
-            this.onResize();
+            // See AKU-775 - adding a short timeout here allows everything to complete initialization before resizing
+            setTimeout(lang.hitch(this, this.onResize), 100);
```
MartinDoyleUK: We're still just guessing that it'll be ready ... perhaps something a little more concrete? There's a more specific solution at http://stackoverflow.com/questions/6287350/how-to-tell-if-tinymce-has-been-initated but even if not that, we could have a while loop that checks every few hundred milliseconds whether it's been loaded yet before continuing?

PR #777, opened 2016-01-18
AKU-779: Film strip view video previews

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-779 to primarily ensure that video previews do not continue to play when they are no longer displayed in the AlfFilmStripView. However, I've taken this further to do some general tidying up of the AlfFilmStripView as well as providing an auto-play feature for the previewer so that it can be configured to automatically play audio/video previews if required.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -655,6 +655,22 @@ define([],function() {
       POST_TO_PAGE: "ALF_POST_TO_PAGE",
 
       /**
+       * This topic should be published when indicating what [previewers]{@link module:alfresco/preview/AlfDocumentPreview}
+       * are currently displayed. This has been added to support the 
+       * [DocumentCarousel]{@link module:alfresco/documentlibrary/views/layouts/DocumentCarousel} so that audio and video
+       * previewers can automatically be stopped and started as they leave are hidden and displayed.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.51
+       *
+       * @event module:alfresco/core/topics~PREVIEWS_SHOWN
+       * @property {object[]} items An array of the items that have been displayed.
+       */
+      PREVIEWS_SHOWN: "ALF_PREVIEWS_SHOWN",
+
+      /**
        * This topic is published to indicate that data needs to be uploaded. This is typically list based data but can
        * be used by other widgets.
        *
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -665,7 +665,7 @@ define([],function() {
        * @default
        * @since 1.0.51
        *
-       * @event module:alfresco/core/topics~PREVIEWS_SHOWN
+       * @event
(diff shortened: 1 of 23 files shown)


aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -655,6 +655,22 @@ define([],function() {
...
+       * are currently displayed. This has been added to support the 
+       * [DocumentCarousel]{@link module:alfresco/documentlibrary/views/layouts/DocumentCarousel} so that audio and video
+       * previewers can automatically be stopped and started as they leave are hidden and displayed.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.51
+       *
+       * @event module:alfresco/core/topics~PREVIEWS_SHOWN
```
MartinDoyleUK: Just `@event`

aikau/src/main/resources/alfresco/preview/AVPlugin.js
```diff
@@ -0,0 +1,98 @@
...
+         this.alfSubscribe(topics.PREVIEWS_SHOWN, lang.hitch(this, this.onPreviewDisplayChange));
+         this.inherited(arguments);
+      },
+
+      /**
+       * Extends the [inherited function]{@link module:alfresco/preview/AlfDocumentPreviewPlugin#onMarkupAdded}
+       * to start playing the preview if [autoPlay]{@link module:alfresco/preview/AVPlugin#autoPlay} is true.
+       * 
+       * @instance
+       * @since 1.0.51
```
MartinDoyleUK: Module has `@since` already, so not need on its members

aikau/src/main/resources/alfresco/preview/AVPlugin.js
```diff
@@ -0,0 +1,98 @@
...
+      },
+
+      /**
+       * Inspects the payload to see if the item represented by this plugin instance is being displayed.
+       * If [autoPlay]{@link module:alfresco/preview/AVPlugin#autoPlay} is true then the preview will begin
+       * playing when displayed, but it will always stop playing when hidden.
+       * 
+       * @instance
+       * @param  {object} payload The list of items that are currently displayed.
+       * @since 1.0.51
```
MartinDoyleUK: Module has `@since` already, so not need on its members

aikau/src/test/resources/alfresco/documentlibrary/views/MediaFilmStripViewTest.js
```diff
@@ -0,0 +1,138 @@
...
+         beforeEach: function() {
+            browser.end();
+         },
+
+         // By default autoPlay is disabled, but the test page overrides the plugin configuration...
+         "Check that video is playing": function() {
+            return browser.findDisplayedByCssSelector("video")
+               .execute("return document.querySelector('video').paused")
+               .then(function(paused) {
+                  assert.isFalse(paused, "The video is not playing");
```
MartinDoyleUK: None of these extra comments are needed. If the test fails you'll get "Check that video is playing" expected true to be false ... really don't believe that extra string is necessary (in the interests of brevity of code if it doesn't impact on meaning).

PR #779, opened 2016-01-19
Add support for synchronous Xhr requests

yregaieg:
as explained in the official dojo documentation: https://dojotoolkit.org/reference-guide/1.7/quickstart/ajax.html#xhr-options dojo/_base/xhr does have an option for Synchronous requests


draperd:
Thanks for providing this change... under normal circumstances it would be good to have a unit test included to ensure that the additional option works. But more importantly (from your point of view) including a unit test will prevent regressions (such as the option being changed or removed). However, in this particular instance I don't mind pulling as it is.... 

Let me know if you'd like to add a unit test, otherwise I'll pull the request later today and it will be included in the 1.0.53 release next Tuesday.

yregaieg:
Thank you for the heads up Dave!
I'll try to create the necessary test within the next few hours.

-----Message d'origine-----
De : "Dave Draper" <notifications@github.com>
Envoyé : ‎19/‎01/‎2016 13:20
À : "Alfresco/Aikau" <Aikau@noreply.github.com>
Cc : "Younes REGAIEG" <y.regaieg@gmail.com>
Objet : Re: [Aikau] Add support for synchronous Xhr requests (#779)

Thanks for providing this change... under normal circumstances it would be good to have a unit test included to ensure that the additional option works. But more importantly (from your point of view) including a unit test will prevent regressions (such as the option being changed or removed). However, in this particular instance I don't mind pulling as it is.... 
Let me know if you'd like to add a unit test, otherwise I'll pull the request later today and it will be included in the 1.0.53 release next Tuesday.

draperd:
OK, cool... You can find info on unit testing here: https://github.com/Alfresco/Aikau/wiki/Unit%20Testing
Also, at the bottom of the main read.me file are setup instructions for Linux, OSX and Windows that cover testing as well.

yregaieg:
Hey dave, I found out that 
> Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/#synchronous-flag .

Please do not merge this until I give it further review.

draperd:
OK, thanks for letting me know... can you describe the use case you have? Have you tried using promises for async xhr?

PR #786, opened 2016-01-22
AKU-746: Create new UploadService

MartinDoyleUK:
This PR addresses primarily [AKU-746](https://issues.alfresco.com/jira/browse/AKU-746), and creates a new `FileUploadService` that works alongside the `StickyPanel` and the `UploadMonitor` to provide a new way of executing and tracking uploads. The existing regression tests have been run and pass, in addition to a full suite where any errors were corrected. This has not yet been tested in situ in Share.

### DIFF
--- a/aikau/.jshintrc
+++ b/aikau/.jshintrc
@@ -30,6 +30,7 @@
       "module": false,
       "model": false,
       "appContext": false,
+      "JSON": false,
       "PDFJS": false
    },
 
--- a/aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
+++ b/aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
@@ -119,9 +119,9 @@
          border-top-color: #ccc;
          cursor: pointer;
          &__highlighted-label {
+            color: @general-font-color;
             font-family: @bold-font;
             font-weight: 700;
-            color: @general-font-color;
             text-decoration: underline;
          }
          &:first-child {
@@ -168,11 +168,11 @@
    }
    &--has-error {
       border-color: #c00;
-      outline: 1px solid #faa;
+      box-shadow: 0 0 1px 1px #ffc4c4;
    }
    &--focused {
       border-color: #09c;
-      outline: 1px solid #0cf;
+      box-shadow: 0 0 1px 1px #80e5ff;
    }
    &--choices-nowrap {
       .alfresco-forms-controls-MultiSelect__choice__content {
--- a/aikau/src/main/resources/alfresco/services/FileUploadService.js
+++ b/aikau/src/main/resources/alfresco/services/FileUploadService.js
@@ -28,7 +28,7 @@
  * @author Martin Doyle
  * @since 1.0.52
  */
-define(["alfresco/buttons/AlfButton", 
+define(["alfresco/buttons/AlfButton",
         "alfresco/core/CoreXhr", 
         "alfresco/core/ObjectTypeUtils", 
         "alfresco/core/topics", 
@@ -42,7 +42,8 @@ define(["alfresco/buttons/AlfButton",
         "dojo/on", 
         "dojo/promise/all", 
         "service/constants/Default"], 
-        function(AlfButton, CoreXhr, ObjectTypeUtils, topics, AlfDialog, _UploadHistoryMixin, BaseService, array, declare, lang, Deferred, on, all, AlfConstants) {
+        function(AlfButton, CoreXhr, ObjectTypeUtils, topics, AlfDialog, _UploadHistoryMixin, BaseService,
+                 array, declare, lang, Deferred, on, all, AlfConstants) {
 
    // Declare and return the class
    return declare([BaseService, CoreXhr, _UploadHistoryMixin], {
@@ -127,14 +128,15 @@ define(["alfresco/buttons/AlfButton",
 
       /**
        * This state-variable is used to track the current overall progress, and is equal to
-       * the number of queued uploads multiplied by 100. This is reset to zero every time
-       * all of the queued uploads complete.
+       * the number of "in progress" uploads. This is incremented by the number of files in
+       * an upload-request, and reset once all of the files have either been uploaded or
+       * failed to upload.
        *
        * @instance
        * @type {number}
        * @default
        */
-      totalProgressPercent: 0,
+      totalNewUploads: 0,
 
       /**
        * The widget that displays the uploads' progress.
@@ -303,9 +305,6 @@ define(["alfresco/buttons/AlfButton",
 
             // Update the display widget with the details of the file that will be uploaded
             this.uploadDisplayWidget.addInProgressFile(fileId, nextFile);
-
-            // Update the total progress percentage
-            this.totalProgressPercent += 100;
          }
 
          // Start uploads
@@ -477,11 +476,14 @@ define(["alfresco/buttons/AlfButton",
                   }));
                }
 
-               // Update the total progress
-               this.updateAggregateProgress();
+               // Update the total number of in-progress files
+               this.totalNewUploads += filesToUpload.length;
 
                // Start the uploads
                this.createFileUploadRequests(filesToUpload, payload.targetData);
+
+               // Update the total progress
+               this.updateAggregateProgress();
             }));
 
          } else {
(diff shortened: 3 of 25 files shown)


aikau/.jshintrc
```diff
@@ -30,6 +30,7 @@
       "module": false,
       "model": false,
       "appContext": false,
+      "JSON": false,
```
draperd: What does this change do?

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
```diff
@@ -168,11 +168,11 @@
    }
    &--has-error {
       border-color: #c00;
-      outline: 1px solid #faa;
+      box-shadow: 0 0 1px 1px #ffc4c4;
```
draperd: Anything special about these box-shadow additions? Different from the defaults LESS variables? Should these be new LESS variables.

aikau/.jshintrc
MartinDoyleUK: Even though JSON parse/stringify is supported by even IE8, it's not part of ES3, so previously JSHint would warn about using it. This explicitly declares it as a valid global object (false means not modifiable).

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
MartinDoyleUK: Good question. The replaced outline values weren't LESS variables, but perhaps they also should have been. The colours are different to account for the different rendering of the two properties, and were judged to be similar by eye. We could certainly make them LESS variables, although it'd be good to tie them (using LESS algorithms) to the colours of our global outline values.
How do you want to do it?

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * <p>This service can be used to control the uploading of content as well as
+ * the updating the content of existing nodes on an Alfresco Repository.</p>
+ * 
+ * @module alfresco/services/FileUploadService
+ * @extends module:alfresco/services/BaseService
```
draperd: There seems to be a lot of duplication in this module from the UploadService. I think we can do a better job of abstraction of common functions here. I want to avoid any possibility of double maintenance.

MartinDoyleUK: To create `FileUploadService`, I started with the `UploadService` and went through line-by-line, removing the hard tie-in with the dialog and the uploads display widget, and tidying up the uploading logic. If we do anything to avoid duplication, I'd suggest we fix `UploadService` by moving the improved code from `FileUploadService` into it.

draperd: Yes, I agree that we should use a mixin, but as we've already allowed scope creep on the changes I'd like us to make the effort to get this done before merging.

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+       * @instance
+       * @readOnly
+       * @type {Object}
+       * @property {num} ADDED File has been queued for upload
+       * @property {num} UPLOADING File being uploaded
+       * @property {num} FINISHED Uploading has finished
+       * @property {num} FAILURE Uploading failed
+       * @property {num} SUCCESS Uploading succeeded
+       */
+      state: {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+       * Constructs the upload payload object to be added to the fileStore object for each file. 
+       * The object constructed is designed to work with the Alfresco REST service for uploading
+       * documents. This function can be overridden to support different APIs
+       *
+       * @instance
+       * @param {object} file The file being uploaded
+       * @param {object} fileName The name of the file being uploaded
+       * @param {object} targetData Information about where and how to upload the data
+       */
+      constructUploadData: function alfresco_services_FileUploadService__constructUploadData(file, fileName, targetData) {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+         return dfd.promise;
+      },
+
+      /**
+       * Starts the actual upload for a file
+       *
+       * @instance
+       * @param {object} Contains info about the file and its request.
+       */
+      startFileUpload: function alfresco_services_FileUploadService__startFileUpload(fileInfo) {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+         fileInfo.request.open("POST", url, true);
+         fileInfo.request.send(formData);
+      },
+
+      /**
+       * Calculates the overall progress of all the uploads and calls the display widget with the data.
+       *
+       * @instance
+       */
+      updateAggregateProgress: function alfresco_services_FileUploadService__updateAggregateProgress() {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+      },
+
+      /**
+       * Validate the supplied collection of files, sending a notification of any invalid ones, and returning the valid ones.
+       *
+       * @instance
+       * @param {object[]} files The files
+       * @returns {object[]} The valid files
+       */
+      validateFiles: function alfresco_services_FileUploadService__validateFiles(files) {
```
draperd: Suitable for mixin - also, you replaced a separate function with a closure - this was done for good reason. You can't easily extend a closure. A 3rd party might want to change how a file is validated without needing to write the array processing - this should go back to how it was.

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+      },
+
+      /**
+       * <p>Reset the state of the service.</p>
+       *
+       * <p><strong>NOTE:</strong> This does not cancel any in-progress uploads.</p>
+       *
+       * @instance
+       */
+      reset: function alfresco_services_FileUploadService__reset() {
```
draperd: Suitable for mixin - this module would have the default, the UploadService would extend to perform the additional actions

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+
+      /**
+       * Called when an upload "load" event fires. This merely means that the
+       * server has responded though, so it could still be an error.
+       *
+       * @instance
+       * @param {object} fileId The unique identifier of the file
+       * @param {object} evt The completion event
+       */
+      processUploadCompletion: function alfresco_services_FileUploadService__processUploadCompletion(fileId, evt) {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+      /**
+       * This function listens for upload progress events retured from the XMLHttpRequest object and
+       * adjusts the display to give a visual indication of how the upload for the related file is
+       * progressing.
+       * 
+       * @instance
+       * @param {string} fileId The unique id of the file being uploaded
+       * @param {object} evt See [ProgressEvent on MDN]{@link https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent}
+       */
+      onUploadProgress: function alfresco_services_FileUploadService__onUploadProgress(fileId, evt) {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
```diff
@@ -0,0 +1,708 @@
...
+      },
+
+      /**
+       * Create the file-upload requests.
+       *
+       * @instance
+       * @param {object[]} filesToUpload The files to be uploaded
+       * @param {object} targetData The data that identifies where to upload the files to.
+       */
+      createFileUploadRequests: function alfresco_services_FileUploadService__createFileUploadRequests(filesToUpload, targetData) {
```
draperd: Suitable for mixin

aikau/src/main/resources/alfresco/services/FileUploadService.js
draperd: OK, I've started comparing and I can see that what you've done is change some of the code, re-order and give them what you think are better function names - however, the core functionality remains the same - I don't want to have different code doing exactly the same thing in two different files.

We can't change the names, but people aren't going to be seeing them - I took them from the original YUI upload - whilst they might not be the best names - they're only names.

I'd like you to go back through and abstract the common code to a mixin - I'm sure that the actual code changes you've made are fine but I'll verify with some testing in place on Share on Monday.

aikau/src/main/resources/alfresco/forms/controls/css/MultiSelect.css
draperd: You can leave this for now - there is already another JIRA issue for addressing LESS changes.

aikau/src/main/resources/alfresco/services/FileUploadService.js
MartinDoyleUK: Yes, you've explained before why you have things as functions instead of closures.

This is still a function, rather than a closure, however the existing functionality limited one to processing a file individually, rather than looking at all of the files in case one wanted to control what types of files could be uploaded simultaneously. Also, the code now correctly allows a failed-validation message to be put into the display, rather than always saying that it's failed because it's 0kb.

I do not believe this should change back to how it was.

MartinDoyleUK:
I note that a lot of the functionality is very similar between the two, however the problem with the UploadService, and the reason I had to basically rewrite it from scratch, was to do with how tied into the old widgets it was. In doing so, I made several things much simpler and/or changed the way they worked so they would be suitable for use in a modeless window, and it will be at the least difficult to abstract everything you suggest to a mixin. If you still want to do that, then I can make the appropriate changes to the UploadService to enable that, however it would certainly add a day or two to the story, which I thought you were keen to get into this sprint. Given that you are suggesting we refactor a deprecated widget in order to make it share code with its replacement, that doesn't seem the most suitable use of effort.

aikau/src/main/resources/alfresco/services/FileUploadService.js
draperd: It should be possible for a 3rd party to extend the service to just handle verification of individual files (without needing to write a loop). It also had another function that allowed verification of all files (i.e. code that creates a loop).

Both have value.

PR #789, opened 2016-01-25
Fixes for AKU-799 CSS regression

kevinroast:
Fixes for AKU-799 CSS regression.
Also fix/remove debug CSS found on AlfList.

FYI the comment says: // NOTE: If this is changed, the associated paddingAdjustment in the JS needs updating

Actually that's only true for *height* adjustment, I have adjusted the width.


draperd:
We'll take this PR as a starting point, but we want to investigate the existing unit tests to see why they haven't picked up this problem and to see if they can be improved. We also want to run a full regression suite to ensure that there are no failures.

It would be good to go back through the log of changes to the file to understand why the CSS was changed (i.e. what issue it relates to) to be sure that this change is not undoing a change made for a specific reason - it might be necessary to adapt the selectors to accommodate multiple scenarios.

draperd:
I think that the fix we need here is to actually stop the SimplePanel created within the dialog to not attempt to handle overflow (because the dialog body is trying to handle that). We also definitely need to improve the existing unit test to prevent future regressions.

PR #794, opened 2016-01-26
Update Sandpit application following early feedback

draperd:
We got some early feedback relating to the Sandpit application that the previews can take some time to load. It was suggested that switching from debug to production move (so that the JS is minified) might increase performance. I've also tweaked the Preview widget so that there is an obvious message that a preview is being loaded.

I'm also using this PR to make a few other minor changes to get the browser title showing correctly.


aikau/src/main/resources/alfresco/prototyping/Preview.js
```diff
@@ -285,6 +285,7 @@ define(["dojo/_base/declare",
             {
                this.rootPreviewWidget.destroyRecursive(false);
             }
+            this.previewNode.innerHTML = this.message("loading.preview.message");
```
MartinDoyleUK: We should be cautious about ever using `innerHTML` over `textContent`. There's no real harm here given the source, but it's worth mentioning.

PR #798, opened 2016-01-28
AKU-791: AlfDialog - Scrolling when dialog is open shows that part of the page is hidden

MartinDoyleUK:
This PR addresses the page still scrolling in https://issues.alfresco.com/jira/browse/AKU-791. The only way to truly fix this is to remove the scrollbar completely when the dialog is visible, however to avoid the display "jumping" it's also necessary to temporarily add a right-hand margin to the body to compensate for the loss of the scrollbar. This has been tested in a local RM instance and on the Aikau test pages, and a full Aikau test suite has been run successfully.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/DialogService.js
+++ b/aikau/src/main/resources/alfresco/services/DialogService.js
@@ -297,6 +297,7 @@ define(["dojo/_base/declare",
        * @instance
        * @type {number}
        * @default
+       * @since 1.0.53
        */
       _scrollbarWidth: null,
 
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -291,6 +292,15 @@ define(["dojo/_base/declare",
       _activeDialogs: [],
 
       /**
+       * The scrollbar width for this browser-environment
+       *
+       * @instance
+       * @type {number}
+       * @default
```
draperd: Since?

aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -859,10 +870,25 @@ define(["dojo/_base/declare",
...
       },
 
+      /**
+       * Get the scrollbar width for the current browser environment. This is cached
+       * after first retrieval for faster access.
+       *
+       * @instance
+       * @returns {number} The scrollbar width
+       */
+      _getScrollbarWidth: function alfresco_services_DialogService___getScrollbarWidth() {
```
draperd: Since?

PR #800, opened 2016-01-28
Add named window support to NavigationService

igorbga:
Add support to NavigationService so that it opens in a named window if the target is not NEW or CURRENT.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -478,7 +478,7 @@ define([],function() {
        * @event module:alfresco/core/topics~NAVIGATE_TO_PAGE
        * @property {string} url - The URL to navigate to
        * @property {string} [type="PAGE_RELATIVE"] - The type of navigation, either "PAGE_RELATIVE", "CONTEXT_RELATIVE", "FULL_PATH" or "HASH"
-       * @property {string} [target=CURRENT"] - Whether to use the current tab ("CURRENT") or open in a new tab ("NEW")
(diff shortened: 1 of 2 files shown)


draperd:
Could you please update this PR to include a unit test in accordance with the acceptance test criteria please. This will ensure that the capability you want to add will not be regressed accidentally in the future.

draperd:
A couple of other things on this PR...

We want to do our best to have good JSDoc (I appreciate that we need to improve ourselves in this area) but one additional thing that would be good is for the JSDoc for the topic to be updated (see https://github.com/Alfresco/Aikau/blob/86f32f9c5777ccf17c832c6692dcb4232d3a8c11/aikau/src/main/resources/alfresco/core/topics.js#L619). This would mean that the online topics documentation (http://dev.alfresco.com/resource/docs/aikau-jsdoc/topics.html) describing the payload would indicate that named targets could be provided.

Finally (and I wouldn't reject a PR for it) but it would be really good if you could keep brackets consistent with the rest of the file :)

As I said, if you'd rather we make the changes then let me know and I'll raise the appropriate JIRA and add to our backlog.

igorbga:
Sorry about the brackets, didn't notice the style difference.

I don't have much clue about how unit testing is done for Aikau so I think it will be better to raise a JIRA issue. I'll try to look to it when it is done to see if I get how to do it for future PRs.

Will you rise the issue yourself or should I raise it?

draperd:
I've raised https://issues.alfresco.com/jira/browse/AKU-824 to cover this feature request. It's in the provisional plan for the sprint after next, so hopefully you'll have it in a couple of weeks - but as this is small we'll try and pull it into the next sprint if possible.

igorbga:
Great Thanks

PR #810, opened 2016-02-01
Updated the project to run test_bamboo against Browserstack.

jphuynh:


aikau/src/test/resources/config/vm/Config.js
```diff
@@ -32,7 +32,6 @@ define({
     * @type {object}
     */
    urls: {
-      unitTestAppBaseUrl: "http://192.168.56.1:8089"
```
draperd: Isn't this going to break our testing on Vagrant? Why does this line need to be removed?

jphuynh:
Closing this for further fix from Martin.

PR #820, opened 2016-02-03
AKU-805: Sandpit previewing

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-805 to update the Sandpit application so that multiple model previews cannot be rendered for each example by repeatedly clicking the preview button. 

It also refactors the Sandpit application to switch runtime mode to "alfresco" that allows us to override the "service/constants/Default" WebScript module in order to always indicate that we are in debug mode (even if we're not). This results in the JS remaining compressed but allows the DebugLog to display publication/subscription activity.

Note that this required that we move the location of the WebScript and Surf object files - so this PR contains a lot of deletion/addition changes that are just moves without changes.

A unit test has been added for the Preview widget.


aikau/src/main/resources/alfresco/prototyping/Preview.js
```diff
@@ -362,56 +401,50 @@ define(["dojo/_base/declare",
...
+               }
             }
-            else
+            
+            // Build in the i18n properties into the global object...
+            for (var scope in response.i18nMap)
             {
-               // ...but if the scope already exists, then we need to mixin the new properties...
-               lang.mixin(window[response.i18nGlobalObject].messages.scope[scope], response.i18nMap[scope]);
+               if (typeof window[response.i18nGlobalObject].messages.scope[scope] === "undefined")
```
MartinDoyleUK: Feels like `window[response.i18nGlobalObject].messages.scope` should be declared in a variable outside the loop to improve readability?

PR #827, opened 2016-02-04
AKU-812: Commas not valid in NumberSpinner

MartinDoyleUK:
This addresses [AKU-812](https://issues.alfresco.com/jira/browse/AKU-812), and has updated the NumberSpinner to work a lot better with other locales now. It could do with someone testing it manually in other locales, but basic testing seems fine, and the regression tests pass.


draperd:
I found that this wasn't reliable when testing on Firefox because the use of .toFixed would sometimes introduce low order decimal place values (see screenshot on JIRA issue). So I have reworked this and will create a new PR.

PR #829, opened 2016-02-08
AKU-812: Support comma as decimal place for appropriate locales

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-812 to make some updates to the NumberSpinner form control to ensure that a comma can be used as the decimal place in locales as applicable. I've re-worked the original PR (#827) which I found to be unreliable when testing on Firefox (see comments on JIRA ticket).

I've changed the way that we were validating numbers and the decimal place check to lean on the dojo/number capabilities to verify locale appropriate number characters.

Unfortunately, we don't have locale specific testing at the moment - so although we have unit tests for decimal place validation, they are en specific - but I've verified this manually.


aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
```diff
@@ -161,30 +224,59 @@ define(["alfresco/forms/controls/BaseFormControl",
...
+         {
+            // Get the raw value of the number box (for locales that use the "," for a decimal place, then the value will include
+            // the ",")...
+            var textBoxValue = this.wrappedWidget.textbox.value;
+
+            // Parse the number (this will convert the "," to ".")...
+            var parsed = number.parse(textBoxValue);
+            if (isNaN(parsed))
+            {
+               // Can't do anything with a value that isn't a number - it should have been picked up by the "isNumberValidator" anyway
```
davidcognite: Could you throw here, otherwise you'll end up failing & bypassing the error reporting in the catch block?

aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
```diff
@@ -161,30 +224,59 @@ define(["alfresco/forms/controls/BaseFormControl",
...
       isNumberValidator: function alfresco_forms_controls_FormControlValidationMixin__isNumberValidator(validationConfig) {
          var isValid = false;
-         try {
-            var value = this._removeCommasAndSpaces(this.wrappedWidget.textbox.value),
-               isNumber = this._valueIsNumber(value);
-            isValid = isNumber || (this.permitEmpty && value === "");
-         } catch (e) {
+         try 
+         {
+            var textBoxValue = this.wrappedWidget.textbox.value;
```
MartinDoyleUK: No. Nothing should access `textbox.value` outside of the widget. `this.wrappedWidget.get("value")` will give you a number or null, and so should be used to check for a valid number.

aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
```diff
@@ -161,30 +224,59 @@ define(["alfresco/forms/controls/BaseFormControl",
...
-               isNumber = this._valueIsNumber(value);
-               numDecimals = value.indexOf(".") === -1 ? 0 : value.split(".")[1].length;
-            isValid = !isNumber || numDecimals === this.permittedDecimalPlaces;
-         } catch (e) {
+         var isValid = true;
+         try 
+         {
+            // Get the raw value of the number box (for locales that use the "," for a decimal place, then the value will include
+            // the ",")...
+            var textBoxValue = this.wrappedWidget.textbox.value;
```
MartinDoyleUK: Again, this should not be used. See above comment re `get("value")`.

aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
draperd: I disagree here... for the simple reason that if you always return null, then it's not possible to tell whether or not the value is NaN or if it is just blank - which means that it's not possible to perform the permitEmpty test.

Let's also try to not lose sight of the fact that we **need** this to work, and it **does** work.

aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
draperd: Unlike the above we could change to use get("value")... but if we're already doing it somewhere else (which we need to)... then I don't see why we need to change it here. The suggestion is that we trust ourselves to access a property a few dozen lines above in the code, but not in this function.

aikau/src/main/resources/alfresco/forms/controls/NumberSpinner.js
draperd: I'm not sure it's necessary here... this validator is only testing the number of decimal places - that's it's only requirement. It's testing that the permitted number of decimal places has **not been exceeded**. So an invalid number error will have been picked up elsewhere and isn't really pertinent to this validation test (bear in mind that this is just one of a set of validations that are being performed).

PR #842, opened 2016-02-10
Add doHashVarUpdate restriction to updateLoadDataPayload of AlfHashList

spotnape:


aikau/src/test/resources/alfresco/lists/AlfHashListTest.js
```diff
@@ -195,6 +196,37 @@ define(["alfresco/TestCommon",
...
+            .getLastPublish("ALF_RETRIEVE_DOCUMENTS_REQUEST")
+               .then(function(payload) {
+                  assert.notProperty(payload, "path", "There should be no path value in the payload");
+               })
+
+            .findByCssSelector("body")
+               .clearLog()
+               .end()
+
+            .get(goodUrl)
```
draperd: One minor change... could you break this test up... and add "Post Coverage Results (1)" before reloading the page with the good URL....

aikau/src/test/resources/alfresco/lists/AlfHashListTest.js
```diff
@@ -195,6 +196,37 @@ define(["alfresco/TestCommon",
...
+               .findByCssSelector("body")
+               .end()
+
+            .getLastPublish("ALF_RETRIEVE_DOCUMENTS_REQUEST")
+               .then(function(payload) {
+                  assert.propertyVal(payload, "path", "d/e/f", "Path value incorrect");
+               });
          }
       };
```
draperd: ...and then add "Post Coverage Results (2") after the test here.

The reason being that we will get the coverage data for the code updates you've put (if we don't post coverage data before leaving a page then it is lost).

draperd:
Looks good, thanks Richard - just one minor change regarding posting coverage data if possible please and its good to merge.

PR #855, opened 2016-02-16
Update UploadContainer.js to take an array of proxyDragAndDropNode selectors

spotnape:
The UploadContainer.js file has been modified and an example test page generated to demonstrate the new capability. The existing test page continues to function as before. Note that there is no automated testing of the existing feature due to the complexity of DnD and the monitoring of visuals, so I have followed the existing approach with the provision of a model for manual testing.


aikau/src/main/resources/alfresco/layout/UploadContainer.js
```diff
@@ -100,18 +104,40 @@ define(["dojo/_base/declare",
        * @instance
        */
       setDndHighlightDimensions: function alfresco_documentlibrary__AlfDndDocumentUploadMixin__setDndHighlightDimensions() {
-         if (this.proxyDragAndDropNode && query(this.proxyDragAndDropNode).length === 1)
+
+         // proxyDragAndDropNode can be an array of strings or a comma-separated list
+         if (this.proxyDragAndDropNode && !(this.proxyDragAndDropNode instanceof Array))
```
draperd: Use of instanceof Array isn't actually sufficient to determine if something is actually an array (see http://web.mit.edu/jwalden/www/isArray.html). We provide the alfresco/core/ObjectTypeUtils isArray function for accurately determining if something is an array or not. This uses the Dojo polyfill for browsers that don't support Array.isArray (see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)

MartinDoyleUK: `Array.isArray()` is available from IE9. An addition to the `shims.js` file would probably be the most suitable way to fix this lack of support in IE8, as it will be temporarily needed and in a single place to remove easily.

The `ObjectTypeUtils.isArray()` method is merely a wrapper for `dojo/_base/lang.isArray()` which has the following test in it:

```javascript
return it && (it instanceof Array || typeof it == "array");
```

In fact, in their comments for the function, Dojo even have written "Does not work on Arrays created in other windows."

If you need any help updating `shims.js` (although it should be obvious) then please holler. We can then use `Array.isArray()` without worrying about the very unlikely event of handling an array across frames.

spotnape:
Updated as requested. Please get back to me if any further problems.

PR #866, opened 2016-02-24
AKU-837: Initial POC for review - NOT TO BE MERGED

draperd:
**NOT FOR MERGING - FOR REVIEW PURPOSES ONLY**

This PR has been raised in relation to https://issues.alfresco.com/jira/browse/AKU-837 to provide the opportunity for the Aikau development community to review enhancements to the testing that allow 3rd party projects to test Aikau widgets reliably and without regression.

The idea is to provide the CSS selectors used by the Intern tests within Java style properties files (it needs to be Java properties files as we have projects with Java based Selenium testing) that can be referenced by a property key in tests.

This allows the Aikau team to rewrite widget DOM structure and classes without breaking 3rd party tests. In order that potential breaking changes (to 3rd party tests) can be easily detected we need to "eat our own dog food" and use these properties files in the Aikau tests.

To do this we are importing the an NPM package (https://www.npmjs.com/package/properties-reader) that allows us to read the properties files. 

The test selectors will be built via a separate Maven target (not implemented yet) and the published to our Maven repository as a separate artifact. 

In order to make internal Aikau testing easier the TestCommon library is updated to include the "getTestSelectors" and "getTestSelector" functions. 

The idea being that a test would call...

```JAVASCRIPT
ar MultiInputSelectSelectors = TestCommon.getTestSelectors("alfresco/forms/controls/MultiInputSelect");
```
...to get the selectors, and then for each selector required can call...

```JAVASCRIPT
var selector1 = TestCommon.getTestSelector(MultiInputSelectSelectors, "options.loaded.state", ["MULTISELECT_1"]);
```
(note the ability to provide an array of substitution tokens).

Please add any comments or suggestions to this PR to the associated JIRA issue please.


aikau/src/test/resources/alfresco/forms/controls/MultiSelectInputTest.js
```diff
@@ -74,13 +77,17 @@ define([
             //    focused element: none
             //    
             "Controls are fully loaded": function() {
-               return browser.findAllByCssSelector("#MULTISELECT_1 .alfresco-forms-controls-MultiSelect--loaded")
+               // Get the selectors required for this test...
+               var selector1 = TestCommon.getTestSelector(MultiInputSelectSelectors, "options.loaded.state", ["MULTISELECT_1"]);
+               var selector2 = TestCommon.getTestSelector(MultiInputSelectSelectors, "options.loaded.state", ["MULTISELECT_2"]);
```
MartinDoyleUK: It might be more readable (when you're halfway down a page of code) to have something more like:

```javascript
var selectors = {
   control1: TestCommon.getTestSelector(MultiInputSelectSelectors, "options.loaded.state", ["MULTISELECT_1"]),
   control2: TestCommon.getTestSelector(MultiInputSelectSelectors, "options.loaded.state", ["MULTISELECT_2"]),
};
```

Then you could just refer to this object in the code (e.g. `selector.control1`), rather than knowing which `selector1` is, or indeed `selector23`.

spotnape:
This is not a comment on this PR proof or the strategy which seems very sensible. But, bearing in mind that some of the other Alfresco teams are also using Intern (SFS definitely - shortly to be another I believe), it might be useful if some of the features being addressed with Intern for Aikau testing were made available as some sort of package. Npm would probably be my selection of choice bearing in mind the way we all work with Intern.

For example, SFS is using some copied items from the Aikau testing area such as the excellent Reporter written by Martin. Other elements we have modified heavily to suit our own test scenarios and use cases. But the idea of a core utils package that brings in the Reporter and a core TestUtils which includes the selector functions described above would be most useful. I don't think we need a full blown test framework for all as it would potentially be too restrictive or we'd find ourselves writing JIRAs to request framework changes and it's time-consuming enough getting tests written already. Intern is the test framework and it can be implemented for Aikau, or SFS, or any other. But with this new feature a package.json entry that brought in some basic utils against the Aikau library would be excellent.

draperd:
We've already looked into abstracting the test framework to make it consumable (https://issues.alfresco.com/jira/browse/AKU-790) and there is some work that has been going on behind the scenes to make this a reality. IMHO SFS should have pushed for abstraction rather than copying what we had (in order to automatically get framework updates). That said, this effort shouldn't prevent SFS from benefiting from what we do with this issue. NPM is under consideration for providing the abstraction though.

aikau/src/test/resources/alfresco/forms/controls/MultiSelectInputTest.js
```diff
@@ -20,14 +20,102 @@
...
-      "intern/dojo/node!leadfoot/keys"
-   ],
+define(["intern!object",
+        "intern/chai!assert",
+        "alfresco/TestCommon",
+        "intern/dojo/node!leadfoot/keys"],
    function(registerSuite, assert, TestCommon, keys) {
 
+      // Get the selectors for the MultiInputSelect widget...
+      var MultiInputSelectSelectors = TestCommon.getTestSelectors("alfresco/forms/controls/MultiInputSelect");
```
MartinDoyleUK: It's "MultiSelectInput" ;)

draperd: I'm forever getting that wrong :) Good spot

draperd:
Closing in preparation for final PR

PR #870, opened 2016-02-24
AKU-859: compactMode causes console error - Add condition

MartinDoyleUK:
This fixes [AKU-859](https://issues.alfresco.com/jira/browse/AKU-859) by adding a condition around the domNode that doesn't exist in compact mode. Full regression test run.


draperd:
Should there be an associated test with this? When this issue was first raised I wasn't able to find an example in our test pages that used the compact mode? Do we not need to add one?

kevinroast:
Yes I noticed that when attempting to use the property originally - no test was present for this particular case on the associated component test page, so it would be prudent to add one to ensure it does not regress.

MartinDoyleUK:
Happy to do that. Couldn't see any particular way of testing this fix, but could add a general compact-mode test.

draperd:
Yes, that would be good... just to check it doesn't blow up as it is at the moment.

PR #874, opened 2016-02-25
AKU-855: warnIfNotAvailable not working for Dates

MartinDoyleUK:
This addresses issue [AKU-855](https://issues.alfresco.com/jira/browse/AKU-855) whereby warnIfNotAvailable is not working for the Date renderer. Copied code from Property renderer and added appropriate test. Full suite run.


draperd:
We should **never** copy paste code. We should break that code out into it's own function and then call it from both modules.

MartinDoyleUK:
Updated following review

PR #876, opened 2016-02-29
AKU-837: Provide test selectors JAR, update MultiSelectInput to use

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-837 (and is a follow-up to #866) to provide a way in which the CSS selectors used for unit testing can be externalised (they will now be built and provided as a separate test JAR). The MultiSelectInput widget tests have been used to test this as a proof-of-concept.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,6 +149,7 @@
       <module>aikau-archetype</module>
       <module>aikau-additional-languages</module>
       <module>aikau-sandpit-application</module>
+      <module>aikau-test-selectors</module>
    </modules>
 
 </project>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -149,6 +149,7 @@
       <module>aikau-archetype</module>
       <module>aikau-additional-languages</module>
       <module>aikau-sandpit-application</module>
+      <module>aikau-test-selectors</module>
```
MartinDoyleUK: Should this be removed?

PR #883, opened 2016-03-01
AKU-837: Correct the name of the MultiSelectInput selectors properties file

draperd:
This is an update to #876 for https://issues.alfresco.com/jira/browse/AKU-837 to set the correct name for the MultiSelectInput widget properties file


davidcognite:
LGTM

PR #884, opened 2016-03-01
AKU-834: Update to provide opt-in configuration

draperd:
This PR updates the previous PR #865 to address https://issues.alfresco.com/jira/browse/AKU-834 as the previous update meant that the selected files were not displayed in the form control. This would be a regression in previous behaviour so we now create an opt-in configuration to enable the behaviour rather than it being the default


kevinroast:
OK sounds good - I have tested this locally with SFS FileSelect override and it is working as expected when the opt-in configuration is set.

PR #887, opened 2016-03-02
AKU-852: Update UploadMonitor to show accessible error icons

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-852 to change the UploadMonitor so that error messages are replaced with an error icon.


andyhealey26:
If we could slightly tweak all these (3?) error messages it'd be good - 

1. Change "could not" to "couldn't".
2. Change "for the following reason" to "because". Would this work in all situations?

So - "The file ''{0}'' could not be uploaded for the following reason. {1}"

would become - The file ''{0}'' couldn't be uploaded because {1}

draperd:
@andyhealey26 I don't think we can do 2. because it won't work for all locales. Also, server side errors are likely to start with a capital letter (which although we could change programmatically) would be getting into dangerous waters with double-byte encoded character sets. I'm just going to do 1. for the moment - I hope you're ok with that?

andyhealey26:
Sure, sounds good.

PR #891, opened 2016-03-03
AKU-846: Form dialog support for "dialogEnableTopic" for button disabling

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-846 to add support for a new parameter when requesting form dialogs: 'dialogEnableTopic'. When this option is used it will result in all the buttons in the dialog being disabled when data is submitted. The topic is subscribed to in order to re-enable the buttons. Using this new feature requires the developer to ensure that the topic is published when failure occurs so that the buttons can be re-enabled to give the user the opportunity to submit valid data.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/DialogService.js
+++ b/aikau/src/main/resources/alfresco/services/DialogService.js
@@ -138,7 +138,9 @@
  * @property {string} [dialogConfirmationButtonTitle="OK"] - The label for the dialog confirmation button
  * @property {string} [dialogConfirmAndRepeatButtonTitle="OK (and repeat)"] The label for the button indicating the dialog should be repeated
  * @property {string} [dialogCancellationButtonTitle="Cancel"] - The label for the dialog cancellation button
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -661,6 +668,25 @@ define(["dojo/_base/declare",
...
+       * @instance
+       * @param {object} dialog The dialog to enable the buttons for
+       * @since 1.0.58
+       */
+      onFailedSubmission: function alfresco_services_DialogService___onFailedSubmission(dialog) {
+         if (dialog)
+         {
+            when(dialog.getButtons(), lang.hitch(this, function(buttons) {
+               array.forEach(buttons, function(button) {
+                  button.set("disabled", false);
```
MartinDoyleUK: What if there are invalid controls in the form?

draperd: You can't submit a form if it has invalid controls. Therefore this scenario won't occur unless the developer intentionally provides a way to publish the enablement topic from within the dialog (because nothing else will be accessible on the screen). This would be quite a foolish thing for them to do, and we can't code against self-destructive behaviour.

PR #896, opened 2016-03-04
AKU-840: Ensure plugin overrides are placed at beginning of the array.

draperd:
This is a further enhancement to the changes made for https://issues.alfresco.com/jira/browse/AKU-840. The issue that the customer was having was that new plugin conditions were being placed at the end of the condition array and the catch-all "imgpreview" thumbnail condition was being matched before the additional condition. This was resulting in the image previewer being used instead of the video previewer. This change ensures that all new conditions will be evaluated first.

### DIFF
--- a/aikau/src/main/resources/alfresco/preview/AlfDocumentPreview.js
+++ b/aikau/src/main/resources/alfresco/preview/AlfDocumentPreview.js
@@ -378,7 +378,7 @@ define(["dojo/_base/declare",
          else
          {
             // The condition provided does not already exist so add it now...
-            this.pluginConditions.push(condition);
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/preview/AlfDocumentPreview.js
```diff
@@ -378,7 +378,7 @@ define(["dojo/_base/declare",
          else
          {
             // The condition provided does not already exist so add it now...
-            this.pluginConditions.push(condition);
+            this.pluginConditions.splice(0, 0, condition);
```
MartinDoyleUK: Can we use `Array.prototype.unshift` method instead?

PR #897, opened 2016-03-04
AKU-207: Add support for Change Type action

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-207 to add support for the Change Type action.


draperd:
@andyhealey26 This has messages to be reviewed

aikau/src/main/resources/alfresco/services/actions/ChangeTypeService.js
```diff
@@ -0,0 +1,326 @@
...
+      /**
+       * Handles requests to change the type of a node. This will perform an XHR request if necessary to load the full
+       * metadata for the node to be upated in order to retrieve its current type.
+       * 
+       * @instance
+       * @param {object} item The item to perform the action on
+       * @fires module:alfresco/core/topics#GET_DOCUMENT
+       */
+      onChangeTypeRequest: function alfresco_services_actions_ChangeTypeService__onChangeTypeRequest(payload) {
+         if (payload && payload.node && payload.node.type)
```
MartinDoyleUK: These payload attributes (and others in this function) aren't defined in `topics.js` against `CHANGE_TYPE_REQUEST`. Is that intentional?

andyhealey26:
Recommended changes below:

aikau/src/main/resources/alfresco/services/actions/i18n/ChangeTypeService.properties

 +change-type.dialog.title=Change ''{0}'' Type
 +change-type.select.label=New Type
 +change-type.success=''{0}' type successfully changed

Questions (both of these will also want updating):

 +change-type.select.desription=The new type for the file. It is only possible to select a type that is a specialization of the existing type.
- What does "a specialization of the existing type" mean?

 +change-type.failure=Unable to change type of document ''{0}''
- When might this not work? Is there a next step we can offer?

draperd:
@andyhealey26 I took the majority of the strings from the existing implementation in Share (the YUI2 version of handling the action). I did however make up the description as we didn't previously have one.

Specialization of the existing type means that if I have a type "cm:sub-content" that inherits from "cm:content", then I can change the type from "cm:content" to "cm:sub-content" (because it is a more specific or specialized version of that type). You cannot switch to a completely different type, and you cannot switch to a more broad type (so you wouldn't be able to subsequently change from "cm:sub-content" back to "cm:content".

The reasons you can't do this are basically because once metadata defined by the specialized type has been applied to a Node it cannot be removed. Hopefully that all makes sense.

The reason why this might fail is because the configured type hierarchy for the client does not match the actual supported data model - what you see in the UI is **not** derived from the model - it is dependant upon developer configuration getting it right. Not all sub-types are exposed, only what a developer (or deployer/administrator) chooses to expose. But they could actually make a mess of it and.

However, we don't **know** what the cause of the failure is, just that the Repository API tells us that the operation failed.

andyhealey26:
Sure, thanks for all the info. Recommend updating the additional 2 strings to:

+change-type.select.desription=Select a new type for the file. You can only select a type that's based on the current type.

+change-type.failure=We couldn't change the type of ''{0}''. Check with your Alfresco Administrator that the type matches the data model.

PR #899, opened 2016-03-07
Fixed loading order of lightbox 'loading' image into init method

kevinroast:
Fixed loading order of lightbox 'loading' image into init method for LightboxService.

Fixes the issue where the "loading" spinner image is never actually loaded (you can see this on the Aikau sandbox page with clickable thumbnails).

Also removes the need for a cheeky window.load event handler to be injected from the non-amd code. The old Lightbox non-amd code is horrible, and needs a rewrite but for a later time...


draperd:
Thanks for providing this fix.

PR #903, opened 2016-03-08
Chrome filesystem API folder listing parsing added to onDndUploadDrop() handler.

kevinroast:
Added recursive parsing of the directory contents and file lists to retrieve a flattened list of files when a folder is dragged and handled by the onDndUploadDrop() handler, to be passed to the UPLOAD_REQUEST topic.

Each item in the flattened list will have an additional "relativePath" property which can be optionally used by an appropriate API to re-create a folder structure if desired. If the property is ignored (which it is currently by existing uploadservice) then the dragged files will simply be uploaded as a flattened list of files. This is good because it means adding this support has no downside to existing upload implementations.

If the browser does not support the FileSystem API (only Chrome 21+ does) then the code will fallback to identical behaviour as before i.e. only handles files in the standard way.

Tested in Chrome, FireFox and IE11. FireFox and IE continue to ignore the folder drag as expected - Chrome will parse the dragged list.

Added a 'dndMaxFileLimit' property to _AlfDndDocumentUploadMixin as a defensive measure against the possibility of a huge folder drag such as C:\ drive by mistake by a user. By default this limit is zero and ignored to maintain existing behaviour.

See SFS-210

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
@@ -159,6 +159,19 @@ define(["dojo/_base/declare",
        * @since 1.0.42
        */
       dndUploadHighlightTimeout: null,
+      
+      /**
+       * Maximum number of files to allow to be published from a single drag operation. As for some user-agents
+       * drag and drop of potentialy large nested folder structures may be supported - this limit will halt the
+       * publish of the file list if it is larger than the supplied value and throw an error. Specify zero to
+       * indicate no limit - which is the default setting to maintain backward compatibility.
+       * 
+       * @instance
+       * @type {integer}
+       * @default
+       * @since 1.0.58
+       */
+      dndMaxFileLimit: 0,
 
       /**
        * Indicates whether or not the mixing module should take advantage of the drag-and-drop uploading capabilities. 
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
```diff
@@ -159,6 +159,19 @@ define(["dojo/_base/declare",
...
+      /**
+       * Maximum number of files to allow to be published from a single drag operation. As for some user-agents
+       * drag and drop of potentialy large nested folder structures may be supported - this limit will halt the
+       * publish of the file list if it is larger than the supplied value and throw an error. Specify zero to
+       * indicate no limit - which is the default setting to maintain backward compatibility.
+       * 
+       * @instance
+       * @type {integer}
+       * @default
+       * @since 1.0.58
```
draperd: 1.0.58 has now been release, this needs to be 1.0.59

aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
```diff
@@ -625,26 +638,97 @@ define(["dojo/_base/declare",
                   username: null
                };
                var updatedConfig = lang.mixin(defaultConfig, config);
-
-               // Check to see whether or not the generated upload configuration indicates
-               // that an existing node will be created or not. If node is being updated then
-               // we need to generate an intermediary step to capture version and comments...
-               if (updatedConfig.overwrite === false)
+               
+               var walkFileSystem = lang.hitch(this, function alfresco_documentlibrary__AlfDndDocumentUploadMixin__onDndUploadDrop__walkFileSystem(directory, callback, error) {
```
draperd: Normally we try to avoid nesting functions like this as this style means that it's not possible for them to be overridden by the widget that is mixing in the module. It also means that the functions themselves stay shorter which aids readability and also allows JSDoc'ing of the function.

aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
```diff
@@ -625,26 +638,97 @@ define(["dojo/_base/declare",
                   username: null
                };
                var updatedConfig = lang.mixin(defaultConfig, config);
-
-               // Check to see whether or not the generated upload configuration indicates
-               // that an existing node will be created or not. If node is being updated then
-               // we need to generate an intermediary step to capture version and comments...
-               if (updatedConfig.overwrite === false)
```
draperd: I'm a bit concerned here that it looks like we're losing the ability to perform drag-and-drop upload, is this correct? This would be a regression in behaviour.

aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
```diff
@@ -625,26 +638,97 @@ define(["dojo/_base/declare",
...
+                     });
+                     
+                     if (callback.pending === 0)
+                     {
+                        callback(callback.files);
+                     }
+                  }, error);
+               });
+               
+               var addSelectedFiles = lang.hitch(this, function alfresco_documentlibrary__AlfDndDocumentUploadMixin__onDndUploadDrop__addSelectedFiles(files) {
```
draperd: As with "walkFileSystem", could we avoid nesting this function please?

draperd:
Is there anyway to mock test data for this (as we have done with the existing file upload behaviour) to prevent any future regressions?

kevinroast:
>As with "walkFileSystem", could we avoid nesting this function please?
The function is private to the impl of that method - so i would rather not extract it.

kevinroast:
>Is there anyway to mock test data for this (as we have done with the existing file upload behaviour) to prevent any future regressions?
I will see if there is a way to mock structure returned by the filesystem api .

kevinroast:
Updated to @since 1.0.59

draperd:
Discussed the testing and code format with @kevinroast  and have agreed that it is not possible to test and that we'll leave the code in the closure style for now (it *could* potentially be refactored later) but there is no obvious reason why 3rd parties would want to access those functions. Whilst my personal preference is for readability and smaller functions, there is no real value in refactoring at the moment.

PR #910, opened 2016-03-10
AKU-816: Finish JSHint fixes

MartinDoyleUK:
This addresses [AKU-816](https://issues.alfresco.com/jira/browse/AKU-816) and removes almost all of the remaining JSHint errors. A full regression suite has been run successfully.

The following errors remain, and will need addressing at some point.

__`'Alfresco' is not defined`__
This is legacy usage of the `Alfresco` global and is present in the following files:
* `src/main/resources/alfresco/core/JsNode.js`
* `src/main/resources/alfresco/reports/SiteContentReport.js`
* `src/main/resources/alfresco/reports/TopSiteContributorReport.js`
* `src/main/resources/alfresco/services/LightboxService.js`

__`'Promise' is not defined`__
This is usage of ES6 code, which even if we used Babel to transpile to ES5 would still fail on older versions of IE. As it is, this will not work on any version of IE. It is present in the following files:
* `src/main/resources/alfresco/preview/PdfJs/PDFFindController.js`
* `src/main/resources/alfresco/preview/PdfJs/PdfJs.js`

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/DojoRadioButtons.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/DojoRadioButtons.js
@@ -23,6 +23,4 @@
  * @author Dave Draper
  * @deprecated Since 1.0.3 - Use [alfresco/forms/controls/RadioButtons]{@link module:alfresco/forms/controls/RadioButtons} instead
  */
-define(["dojo/_base/declare",
-        "alfresco/forms/controls/RadioButtons"], 
-        function(declare, RadioButtons) {});
+define([], function() {});
--- a/aikau/src/main/resources/alfresco/forms/controls/DojoRadioButtons.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/DojoRadioButtons.js
@@ -23,4 +23,8 @@
  * @author Dave Draper
  * @deprecated Since 1.0.3 - Use [alfresco/forms/controls/RadioButtons]{@link module:alfresco/forms/controls/RadioButtons} instead
  */
-define([], function() {});
+define(["dojo/_base/declare",
+        "alfresco/forms/controls/RadioButtons"],
+        function(declare, RadioButtons) {
+   return declare([RadioButtons], {});
+});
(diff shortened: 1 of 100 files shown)


aikau/src/main/resources/alfresco/forms/controls/DojoRadioButtons.js
```diff
@@ -23,6 +23,4 @@
  * @author Dave Draper
  * @deprecated Since 1.0.3 - Use [alfresco/forms/controls/RadioButtons]{@link module:alfresco/forms/controls/RadioButtons} instead
  */
-define(["dojo/_base/declare",
-        "alfresco/forms/controls/RadioButtons"], 
-        function(declare, RadioButtons) {});
\ No newline at end of file
+define([], function() {});
```
draperd: I don't get this change.

MartinDoyleUK: The callback supplied to the define function is empty, therefore there's no point in loading either declare or RadioButtons into the closure for that function. It's simply a removal of redundant code.

aikau/src/main/resources/alfresco/preview/PdfJs/TextLayerBuilder.js
```diff
@@ -300,24 +302,13 @@ define(["dojo/_base/declare",
             offset: undefined
           };
 
-          function beginText(begin, className) {
```
MartinDoyleUK: These two functions were moved down because they both call a function that was previously defined after them.

aikau/src/main/resources/alfresco/services/PreviewService.js
```diff
@@ -50,18 +50,20 @@ define(["dojo/_base/declare",
...
       generateDependencies: function alfresco_services_PreviewService__generateDependencies(payload) {
+         /*jshint eqnull:true*/
          if (payload != null && payload.model != null)
          {
-            var pageDefObject = dojoJson.parse(payload.model);
-            var data = {
-               jsonContent: pageDefObject,
-               widgets: payload.model
-            };
+            // This block (and query below) commented out because JSHint said they were unused .... why unused though? Why are they here?
```
draperd: Code probably got copied somewhere else during a refactor. Happy to leave this comment in the code for now and investigate at a later date

aikau/src/main/resources/alfresco/forms/controls/DojoRadioButtons.js
draperd: Ah, I think I've actually made a mess of this deprecation. Compare with the changes I made to DojoSelect for example... DojoRadioButtons should be inheriting from RadioButtons now - could you correct this to actually do so please.

PR #913, opened 2016-03-11
AKU-873: AlfMenuBarSelect (and related fixes)

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-873 to ensure that icon updating works correctly with the AlfMenuBarSelect. However, along the way I've fixed up the basic menus tests and ensured that popup menus can share the same icon suite as other menu widgets.


aikau/src/main/resources/alfresco/menus/AlfMenuBarPopup.js
```diff
@@ -109,32 +109,51 @@ define(["dojo/_base/declare",
...
             this.title = this.message(this.title);
          }
+         if (!this.iconAltText && this.label)
+         {
+            this.iconAltText = this.label;
+         }
+         else
+         {
+            this.iconAltText = this.message(this.iconAltText);
+         }
```
MartinDoyleUK: It looks like `this.label` won't be "messaged"? Could this just be written as `this.iconAltText = this.message(this.iconAltText || this.label);`?

aikau/src/main/resources/alfresco/menus/_AlfMenuItemMixin.js
```diff
@@ -148,13 +149,24 @@ define(["dojo/_base/declare",
       postMixInProperties: function alfresco_menus__AlfMenuItemMixin__postMixInProperties() {
          if (this.label)
          {
-            this.params.label = this.label = this.encodeHTML(this.message(this.label));
+            this.params.label = this.label = this.message(this.label);
          }
-         if (this.title)
+         if (!this.title)
+         {
+            this.title = this.label;
```
MartinDoyleUK: Again, a possible non "messaging"??

aikau/src/main/resources/alfresco/menus/AlfMenuBarPopup.js
draperd: It's not obvious from how Git shows the diff, but this.label has already been "messaged" directly before, so if it's available it's the messaged form.

aikau/src/main/resources/alfresco/menus/_AlfMenuItemMixin.js
draperd: As before, except I'm missing the check that label is set

PR #923, opened 2016-03-15
AKU-869: Updates to NodePreviewService following manual testing

draperd:
This PR makes some further updates in relation to https://issues.alfresco.com/jira/browse/AKU-869 that update the preview PR #919 . An additional test has been added to verify the code path that was not previously being used.

PR #925, opened 2016-03-15
AKU-869: Update process function order to prevent errors

draperd:
This PR is a minor tweak to #923 in relating to https://issues.alfresco.com/jira/browse/AKU-869 to prevent errors when processing currentItem tokens **after** setting the currentItem for custom previewing via the NodePreviewService


kevinroast:
Tested this locally with Previewer changes.

Issue #904, opened 2016-03-09
FilteringSelect and MultiSelectInput form controls do not show any options to use


aviriel:
Aikau 1.0.39.5 + Alfresco 5.1.e.

I try to add a text field with autocomplete (or something like that) to the form dialog to select users.

The following control works good:
```
{
	name: "alfresco/forms/controls/Select",
	config: {
		fieldId: "select1",
		label: "Select a person",
		name: "select1",
		optionsConfig: {
			publishTopic: "ALF_GET_FORM_CONTROL_OPTIONS",
			publishPayload: {
				url: AlfConstants.PROXY_URI + "api/people",
				itemsAttribute: "people",
				labelAttribute: "userName",
				valueAttribute: "userName"
			}
		}
	}
}
```
So I suggest that FilteringSelect and MultiSelectInput should work in similar way, but they don't:
```
{
	name: "alfresco/forms/controls/FilteringSelect",
	config: {
		fieldId: "people1",
		name: "people1",
		label: "People",
		optionsConfig: {
			queryAttribute: "value",
			publishTopic: "ALF_GET_FORM_CONTROL_OPTIONS",
			publishPayload: {
				url: AlfConstants.PROXY_URI + "api/people",
				itemsAttribute: "people",
				labelAttribute: "userName",
				valueAttribute: "userName"
			}
		}
	}
}, {
	name: "alfresco/forms/controls/MultiSelectInput",
	config: {
		label: "people2",
		name: "people2",
		optionsConfig: {
			queryAttribute: "label",
			publishTopic: "ALF_GET_FORM_CONTROL_OPTIONS",
			publishPayload: {
				url: AlfConstants.PROXY_URI + "api/people",
				itemsAttribute: "people",
				labelAttribute: "userName",
				valueAttribute: "userName"
			}
		}
	}
}
```
FilteringSelect does not show any suggestions and always shows "The value entered in not valid", but request is sent and correct response is returned from the repository
![selection_108](https://cloud.githubusercontent.com/assets/596948/13632042/d4db8310-e5f8-11e5-801e-03b13f99c407.png)

MultiSelectInput does not even sending request and shows "Loading" all the time:
![selection_109](https://cloud.githubusercontent.com/assets/596948/13632227/aeae2822-e5f9-11e5-8c35-9633446e88ab.png)

Any ideas what to do with it? Usually I think that I do something incorrectly, but in this time Axel Faust said that it may be a bug. I copy the discussion from IRC here, maybe it can help:
[...]

PR #930, opened 2016-03-16
AKU-884: Ensure OptionsService works with ServiceStore

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-884 (and #904) to ensure that form controls that use the ServiceStore (such as FilteringSelect) and those that don't (such as Select) can both make use of the OptionsService. Full unit tests have been added for the use case provided which related to retrieving user information from the repository. In addition, the UserService has been updated to provide the capability of returning user data with the "displayName" attribute that is missing from the REST API.


aikau/src/main/resources/alfresco/core/topics.js
```diff
@@ -652,6 +652,21 @@ define([],function() {
       GET_PREFERENCE: "ALF_PREFERENCE_GET",
 
       /**
+       * This topic can be published to request a list of all available users in the Alfresco Repository.
+       * This request is expected to be handled by the [UserService]{@link module:alfresco/services/UserService}.
+       * The results are published on 'alfResponseTopic' provided in the attribute "items".
```
MartinDoyleUK: This block all made sense apart from the final 5 words. Can you explain better? Where is this attribute found?

aikau/src/main/resources/alfresco/forms/controls/MultiSelect.js
```diff
@@ -1371,8 +1371,8 @@ define([
             this._focusedResult = null;
 
             // If we have too many current results displaying, remove the excess nodes
-            while (this._nodes.resultsDropdown.childNodes.length > this._results.length + 3) {
-               this._nodes.resultsDropdown.removeChild(this._nodes.resultsDropdown.lastChild);
+            while (this._nodes.resultsDropdown.childNodes.length > 3) {
```
MartinDoyleUK: No. This is wrong. You should be removing excess results, not all results. The idea is to avoid re-generating DOM elements for every result if they're already there and can simply have their text content updated.

aikau/src/main/resources/alfresco/services/OptionsService.js
```diff
@@ -109,7 +121,22 @@ define(["dojo/_base/declare",
...
+
+         // See AKU-884...
+         // The topic that the options are currently published on is tactically chosen to only use the
+         // alfResponseScope in tandem with the alfResponseTopic - but for the responseScope to be used by itself
+         // this is to retain backwards compatibility with form controls that do not use the ServiceStore (such
+         // as select) and those that do (such as FilteringSelect)...
+         var topic;
+         if (originalRequestConfig.data.alfResponseTopic)
+         {
+            topic = (originalRequestConfig.data.alfResponseScope || "") + originalRequestConfig.data.alfResponseTopic;
```
MartinDoyleUK: We shouldn't be manually scoping topics if we can avoid it ... is there no way of providing the scope when doing the publish?

aikau/src/main/resources/alfresco/services/UserService.js
```diff
@@ -64,12 +66,91 @@ define(["dojo/_base/declare",
...
+      },
+
+      /**
+       * Handles requests to retrieve all users from the Alfresco Repository.
+       * 
+       * @instance
+       * @param  {object} payload The payload for the request
+       * @since 1.0.60
+       */
+      onGetUsers: function alfresco_services_UserService__onGetUsers(/*jshint unused:false*/ payload) {
```
MartinDoyleUK: `payload` is used

aikau/src/main/resources/alfresco/services/OptionsService.js
draperd: I'd have thought the massive comment above would have explained why this has to be done :)

aikau/src/main/resources/alfresco/forms/controls/MultiSelect.js
draperd: I've tried re-instating that line, but it doesn't work.

aikau/src/main/resources/alfresco/services/OptionsService.js
MartinDoyleUK: Sorry. Lots of words. Hard to read :)

PR #931, opened 2016-03-16
AKU-895: AlfBreadcrumbTrail doesn't reflect hash on load

MartinDoyleUK:
This addresses issue [AKU-895](https://issues.alfresco.com/jira/browse/AKU-895) by ensuring that, when loaded, `AlfBreadcrumbTrail` will read the hash and apply it. It also adds hash support for filters as opposed to just paths. Tests have been added and a full test suite run successfully. Additionally, a minor bug in `urlUtils` has been fixed.


aikau/src/main/resources/alfresco/documentlibrary/AlfBreadcrumbTrail.js
```diff
@@ -268,21 +269,22 @@ define(["dojo/_base/declare",
                this.alfSubscribe(this.metadataChangeTopic, lang.hitch(this, this.onCurrentNodeChange));
             }
 
+            if (this.currentPath)
+            {
+               this.renderPathBreadcrumbTrail();
+            }
```
MartinDoyleUK: This block not new, just moved above the hash handlers to ensure default state loaded before hash applied.

Issue #908, opened 2016-03-09
Incorrect locale sent in Accept-Language header (in IE11)


AFaust:
https://issues.alfresco.com/jira/browse/AKU-612 introduced an attempt to manually derive the "Accept-Language" header from the navigator object. Unforunately, this breaks / overrides the correct Accept-Language header already set by some browsers (like IE11) and may introduce incorrect values.
I.e. my system has been installed in locale de_DE but I usually set the browser content language to just en or de. Regardless of what I set in IE 11, Aikau XHR requests always send de-de in Accept-Language. If I override the getDefaultHeaders in my services to delete the custom-constructed value, the Accept-Language header is set by IE and correct.

As far as I could determine from https://msdn.microsoft.com/en-us/library/ms535867%28v=vs.85%29.aspx there is no property in IE navigator object to determine the correct content language as defined by the user. I would prefer CoreXhr not try to "guess" the header value if the browser is already providing it. Otherwise it should be fixed - the simplest (and most likely to be correct) option seems to be to inject the full Accept-Language header value provided in the initial request to the Aikau page as the default value.

This is using Aikau 1.0.51 on Alfresco 5.0.d

![aikau-xhr-acceptlanguage](https://cloud.githubusercontent.com/assets/2211521/13653541/982b6a7a-e652-11e5-8f36-eafec24e21ab.png)

draperd:
I've raised https://issues.alfresco.com/jira/browse/AKU-885 to address this issue. As this is a bug it will automatically be included in the next sprint, hopefully this will therefore be fixed in the 1.0.60 release.

draperd:
Hopefully you've seen the comments added to the JIRA ticket. I'm not sure how best to resolve this issue as the Accept-Language header is being set appropriately in accordance with the operating system language (if not the browser language). It doesn't appear to be possible to obtain the browser locale via JavaScript (either via `navigator.userLanguage` or `clientInformation.browserLanguage` (which always return the OS language).

The question is how common is this requirement going to be to the typical end user (as opposed to a developer). Our priority is toward not regressing AKU-612 and to ensure optimum performance. The only option to accurately obtain the browser configuration on IE is to fire and a dummy XHR request in order to obtain the Accept Language header set by IE and then use that in the following request.

Unless I can determine that all versions of IE are setting the Accept Language header on POST (the issue reported for AKU-612) in which case we can skip the header setting just for IE (I've not investigated this yet).

draperd:
[...]

PR #934, opened 2016-03-17
AKU-885: Ensure correct Accept-Language header on IE

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-885 (and #908) to ensure that the Accept-Language header is set for Internet Explorer. No unit tests have been added (because we don't yet have IE testing running automatically yet), but manual testing has been carried out against BrowserStack.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/CoreXhr.js
+++ b/aikau/src/main/resources/alfresco/core/CoreXhr.js
@@ -431,12 +431,18 @@ define(["dojo/_base/declare",
          if (languages) 
          {
             languages = array.map(languages, function(nextLang) {
-               return nextLang.toLowerCase();
+               return nextLang;
             }).join(", ");
          } 
          else 
          {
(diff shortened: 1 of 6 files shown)


aikau/src/main/resources/alfresco/core/CoreXhr.js
```diff
@@ -427,12 +428,15 @@ define(["dojo/_base/declare",
...
          // Build up the Accept-Language header value
          var languages = navigator.languages;
-         if (languages) {
+         if (languages) 
+         {
             languages = array.map(languages, function(nextLang) {
                return nextLang.toLowerCase();
             }).join(", ");
-         } else {
-            languages = (navigator.language || navigator.userLanguage).toLowerCase();
```
MartinDoyleUK: Can't help wondering if we should leave this as a fallback (until we've done more cross-browser testing) to make sure this is always used as a last resort?

PR #935, opened 2016-03-18
Improvements to folder DnD FileSystem API handling of many entries per folder.

kevinroast:
It is a commonly held misconception that the DnD FileSystem API has a max limit of 100 items that can be processed per folder level:
See https://productforums.google.com/forum/#!topic/drive/8c-9XvbZDTE
Also https://github.com/owncloud/core/issues/19834

However! It appears it is in reality a slightly esoteric API pattern at fault, not a limitation as per the MDN docs on the subject:
"Continue calling readEntries() until an empty array is returned. You have to do this because the API might not return all entries in a single call."

Refactored the handling to perform the above change. Only affects Chrome folder DnD processing, and has been manually tested.


vuRoland:
since topic was merged with a similar, i'll post the same info for a possible fix for this bug: 

https://github.com/blueimp/jQuery-File-Upload/commit/911dcf1f5bbfda877573fd4fb2b67dd47f225752

This seems to fix the >100 files in a folder, drag-drop issue in Chrome.

PR #936, opened 2016-03-18
AKU-879: Improve DND upload overlay positioning

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-879 to improve the way in which drag and drop upload overlays are handled. The position and dimensions of the overlay are now controlled so that the overlay is constrained to within the view port.

It's not possible automate unit tests for this, however a new unit test page has been provided to manually verify all the various scroll states with both full window scrolling and nested element scrolling.

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
@@ -45,9 +45,11 @@ define(["dojo/_base/declare",
         "dojo/dom-geometry",
         "dojo/dom-style",
         "dojo/dom",
-        "dojo/_base/window"], 
+        "dojo/_base/window",
+        "jquery",
+        "jqueryui"], 
         function(declare, AlfCore, _AlfDocumentListTopicMixin, PathUtils, topics, lang, array, mouse, on, registry, domClass, 
-                 domConstruct, domGeom, domStyle, dom, win) {
+                 domConstruct, domGeom, domStyle, dom, win, $) {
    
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/documentlibrary/_AlfDndDocumentUploadMixin.js
```diff
@@ -579,16 +581,144 @@ define(["dojo/_base/declare",
...
+       */
+      findScrollParent: function alfresco_documentlibrary__AlfDndDocumentUploadMixin__findScrollParent(domNode) {
+         var scrollParent = $(domNode).scrollParent();
+         if (!scrollParent.is("html") &&
+             scrollParent[0].clientHeight === scrollParent[0].scrollHeight)
+         {
+            scrollParent = this.findScrollParent(scrollParent[0]);
+         }
+         return scrollParent;
+      },
```
MartinDoyleUK: This doesn't check that the element can scroll, only that it's not displaying its entire content. Have a look at the `hasScrollbars()` method in `TestCommon` to see a more definitive test. I also worry that a scrollHeight/contentHeight equality test might occasionally fail because the content is too small for a container? Not 100% sure about that though.

draperd: I'm not sure you're correct here, see the documentation for the function being used: https://api.jqueryui.com/scrollParent/

The behaviour is working as expected with the test page that executes the code.

MartinDoyleUK: Had a look at the source code, and it still doesn't seem as comprehensive, but if it's working then I guess it's moot. Merging.

PR #946, opened 2016-03-21
AKU-894 - Make it possible to submit a form dialogue using the enter key

spotnape:
Following a pattern defined by Martin, this modification adds the same functionality to form dialogs when configured. A new test is included. Documentation changes to DialogService.js included.


draperd:
Please nobody merge this yet. We need to have a discussion on the approach to be taken.

PR #949, opened 2016-03-22
AKU-887: Add Appendix even if there is no data.

davidcognite:
Attached code & start of test. As discussed, the test model doesn't render as expected - the view fails to load.


draperd:
Closing PR as discussed. We established that the issue was that the appendix widgets need to match the structure of the view. I'll add more information to the associated JIRA ticket

PR #954, opened 2016-03-24
Update Surf version

draperd:
This PR makes updates to start using Surf 6.3 for testing and in the clients created via the archetype. The test application is also updated to take advantage of faster external LESS pre-processing via Node. This requires an NPM update. Further updates have been made in relation to versioned WebScripts to ensure that multiple Aikau JARs can co-exist without WebScript collisions. In a previous PR it was ensured that WebScript URLs would be different - but this change ensures that they are in different locations so that they can be loaded and processed.


aikau/package.json
```diff
@@ -1,29 +1,30 @@
...
+    "jshint-stylish": "2.1.0",
+    "less": "^2.6.1",
+    "load-grunt-tasks": "3.4.1",
+    "node-coverage": "1.0.7",
+    "properties-reader": "0.0.15",
+    "safe-json-serialiser": "0.0.3",
+    "tcp-port-used": "0.1.2",
+    "time-grunt": "1.3.0"
+  }
+}
```
MartinDoyleUK: Not keen on the total whitespace reformat here that NPM does automatically. Will put it back to 3-spaces next time I update the dependencies, so no change needed at the moment.

PR #965, opened 2016-03-30
AKU-908: Convert inline edit actions labels to buttons

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-908 to swap the action labels in the InlineEditProperty (and descendant widgets) with the previously hidden form buttons. The unit tests have been updated and the InlineEditProperty and InlineEditSelect test selectors have been externalised.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/css/Form.css
+++ b/aikau/src/main/resources/alfresco/forms/css/Form.css
@@ -6,4 +6,13 @@
    .alfresco-forms-Form__warnings--hidden {
       display: none;
    }
+
+   &--single-line {
+      >div {
+         display: inline-block;
+      }
+      >form {
+         display: inline-block;
+      }
+   }
 }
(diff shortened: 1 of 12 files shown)


aikau/src/main/resources/alfresco/forms/css/Form.css
```diff
@@ -6,4 +6,13 @@
    .alfresco-forms-Form__warnings--hidden {
       display: none;
    }
+
+   &--single-line {
+      >div {
+         display: inline-block;
+      }
+      >form {
```
MartinDoyleUK: Could be `> div, > form {` but makes no difference here.

aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
```diff
@@ -369,19 +365,27 @@ define(["dojo/_base/declare",
...
        */
       getFormWidget: function alfresco_renderers_InlineEditProperty__getFormWidget() {
-         if (this.formWidget === null)
+         if (!this.formWidget)
          {
             var uuid = this.generateUuid();
-            this.alfSubscribe(uuid + "ALF_FORM_VALIDITY", lang.hitch(this, this.onFormValidityChange), true);
+            var saveTopic = "_SAVE";
+            var cancelTopic = "_CANCEL";
+            this.alfSubscribe(uuid + saveTopic, lang.hitch(this, this.onSave), true);
```
MartinDoyleUK: Would be good refactor this at some point to allow specification of a scope in the arguments, rather than having to do string concatenation. Perhaps a candidate for 1.1.0?

PR #967, opened 2016-03-31
AKU-904: Add support for additional form configuration in dialogs

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-904 to add support for additional form configuration to be applied to form dialogs. This PR also fixes a Firefox specific test failing.


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -151,6 +151,8 @@
  * @property {boolean} [fullScreenMode=false] Whether or not to create the dialog the size of the screen
  * @property {boolean} [fullScreenPadding=10] The padding to leave around the dialog when in full screen mode
  * @property {boolean} [showValidationErrorsImmediately=true] Indicates whether or not to display form errors immediately
+ * @property {object} [customFormConfig=null] Any additional configuration that can be applied to a [Form]{@link module:alfresco/forms/Form} (please note that the following form configuration
```
MartinDoyleUK: I don't think we need to specify a default of `null` on an optional object property in the JSDoc (not that it makes any difference, and so I'll merge regardless of this).

PR #970, opened 2016-04-01
AKU-909: Update actions renderer

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-909 by re-implementing the alfresco/renderer/Actions and alfresco/renderer/XhrActions widgets to prevent the issue occurring (and by doing so simultaneously addressing https://issues.alfresco.com/jira/browse/AKU-772 and https://issues.alfresco.com/jira/browse/AKU-850). Unit tests have been updated and test selectors externalised.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
@@ -316,6 +316,64 @@ define(["dojo/_base/declare",
          {
             this.alfLog("log", "Skipping action as it's missing from whitelist: " + action);
          }
+      },
+
+      /**
+       * 
+       * @instance
+       * @since 1.0.62
+       */
+      createEmptyMenu: function alfresco_renderers__ActionsMixin__createEmptyMenu() {
+         this.dropDown = this.createWidget({
+            id: this.id + "_DROPDOWN",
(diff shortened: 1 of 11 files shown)


aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
```diff
@@ -316,6 +316,64 @@ define(["dojo/_base/declare",
...
+      /**
+       * 
+       * @instance
+       * @param  {Function} callback The function to call to display the drop-down
+       * @since 1.0.62
+       */
+      createDropDownMenu: function alfresco_renderers__ActionsMixin__createDropDownMenu(callback) {
+         this.createEmptyMenu();
+         this.addActions();
+         callback();
```
MartinDoyleUK: Is there a reason for changing to the callback paradigm than using promises as we tend to elsewhere?

draperd: Yes, because we're hooking into the way that Dojo provides the drop-down handling capabilities, see https://dojotoolkit.org/reference-guide/1.10/dijit/_HasDropDown.html for the details on the mixin we're leveraging and the recommendations for asynchronous loading of the menu.

PR #976, opened 2016-04-06
AKU-916: Dialog scrolls page

MartinDoyleUK:
This PR addresses [AKU-916](https://issues.alfresco.com/jira/browse/AKU-916) where opening a dialog scrolls the page. Have changed AlfDialog and updated associated tests. Also reset defaults in Suites.js

### DIFF
--- a/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
+++ b/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
@@ -62,7 +62,31 @@ define(["dojo/_base/declare",
         function(declare, Dialog, AlfCore, CoreWidgetProcessing, ResizeMixin, topics, _FocusMixin, lang, sniff, array,
                  domConstruct, domClass, domStyle, domGeom, html, aspect, on, when, $) {
    
-   return declare([Dialog, AlfCore, CoreWidgetProcessing, ResizeMixin, _FocusMixin], {
+    /**
+    * This is a customization of the default dijit/Dialog implementation to support
+    * AKU-916, preventing dialogs from scrolling the page when opened.
+    */
+   var CustomDialog = declare([Dialog], {
+
+      /**
+       * Override the default dialog method to ensure that the dialog starts its position
+       * at the top of the page to avoid the page scrolling to focus on its content (see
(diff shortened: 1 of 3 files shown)


aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
```diff
@@ -62,7 +62,31 @@ define(["dojo/_base/declare",
         function(declare, Dialog, AlfCore, CoreWidgetProcessing, ResizeMixin, topics, _FocusMixin, lang, sniff, array,
                  domConstruct, domClass, domStyle, domGeom, html, aspect, on, when, $) {
    
-   return declare([Dialog, AlfCore, CoreWidgetProcessing, ResizeMixin, _FocusMixin], {
+    /**
+    * This is a customization of the default dijit/Dialog implementation to support
+    * AKU-916, preventing dialogs from scrolling the page when opened.
+    */
+   var CustomDialog = declare([Dialog], {
```
draperd: I have to admit - I don't actually understand why we need to create a custom dialog module and then extend it. Give that we were originally just extending Dialog, why not just extend the `show` method directly? Why the indirection? It feels unnecessary.

aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
```diff
@@ -62,7 +62,31 @@ define(["dojo/_base/declare",
...
+       * call to child.focus() below).
+       * 
+       * @instance
+       * @override
+       * @returns {Promise} Returns the superclass' promise
+       * @since 1.0.63
+       */
+      show: function alfresco_dialogs_AlfDialog__CustomDialog__show() {
+         domStyle.set(this.domNode, {
+            top: "0"
```
draperd: What happens if the page is already partially scrolled? Does this mean that opening a dialog always returns to the top? If this is the case (and the JSDoc suggests it is) then this fix feels like a bug waiting to be raised.

MartinDoyleUK:
_NOTE: Scroll pos gained from this Stackoverflow code - http://stackoverflow.com/a/16581384/360424_

PR #979, opened 2016-04-07
AKU-922: Service failure handling and lightbox updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-922 to ensure that the root Page module only decrements the registered services counters once when a service throws an exception on initialization. Updates have also been made to ensure that the LightBoxService will now always load the version of ligthbox.js from the same Aikau JAR. A unit tests has been added to prevent regressions in page behaviour.

### DIFF
--- a/aikau/src/test/resources/alfresco/core/PageTest.js
+++ b/aikau/src/test/resources/alfresco/core/PageTest.js
@@ -1,0 +1,40 @@
+/**
+ * Copyright (C) 2005-2016 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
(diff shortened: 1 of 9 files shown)


aikau/src/test/resources/alfresco/core/PageTest.js
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * This test assesses the CoreRwd mixin as applied to AlfMenuBarPopup
+ *
+ * @author Richard Smith
+ * @author Dave Draper
+ */
```
MartinDoyleUK: File seems new, but no @since?

aikau/src/test/resources/testApp/js/aikau/testing/services/ThrowsInitException.js
```diff
@@ -0,0 +1,41 @@
...
+   
+   return declare([BaseService], {
+      
+      /**
+       * Intentionally causes an exception for testing purposes
+       * 
+       * @instance
+       */
+      initService: function aikauTesting_services_ThrowsInitException__initService() {
+         this.willSurelyDoIt();
```
MartinDoyleUK: How about `throw new Error("This is an error");`? The current code looks a tad arbitrary (like maybe it's a function defined elsewhere??

draperd: You're no fun.

aikau/src/test/resources/alfresco/core/PageTest.js
draperd: It's a test page, so won't show up in the JSDoc, but it probably won't hurt to add one. Do we want to be doing this for all files now (purely for the benefit of the developer editing the file)?

MartinDoyleUK: Fair point .... hadn't noticed that. Perhaps we should try and JSDoc all our files, even the test ones, as a means of documenting them for the next developer?

PR #980, opened 2016-04-07
AKU-914: Add widgets to notifications

MartinDoyleUK:
This addresses [AKU-914](https://issues.alfresco.com/jira/browse/AKU-914) by adding the ability to configure widgets in notifications. A unit-test has been added and a full regression suite run successfully.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/topics.js
+++ b/aikau/src/main/resources/alfresco/core/topics.js
@@ -342,6 +342,11 @@ define([],function() {
        *
        * @event
        * @property {string} message The message to be displayed
+       * @property {string} [autoClose=true] Whether the notification should automatically close itself
+       * @property {object[]} [widgets] Widgets to be inserted into the notification below the message. Note that
+       *                                any widgets required must be either statically declared in the page model
+       *                                or in a widgets property of a custom widget/service, or be specifically
+       *                                required by a custom widget/service.
        * @property {string} [publishTopic] A topic to be published after the notification has closed
        * @property {object} [publishPayload] The payload to be published after the notification has closed
        * @property {boolean} [publishGlobal] Whether to publish the topic globally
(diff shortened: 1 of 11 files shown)


draperd:
My only comment on this is that if you compare with the screenshot in https://issues.alfresco.com/jira/browse/SFS-375 then you'll see that the action is horizontally aligned to the right of the message, where as in the unit test you've created the button is appearing below. It might be worth running this by the SFS team for an opinion on this - it might be better for the the widgets position to be right aligned - at least initially, maybe in the future we could add configuration options for placement.

MartinDoyleUK:
Have implemented it as discussed with Richard (and that was one of the things discussed). That's one of the reasons I added a `notificationId` payload property in, so it would be easier for him to style the notification and its contents accordingly. He was happy to apply CSS to fix any layout requirements they may have.

draperd:
Hmmm... not sure about that. We don't support CSS overlays - great for SFS for now, but this then goes out into the public domain and becomes the default which to me feels sub-optimal - if it's a better design to have the widgets right aligned then that should be the default, we shouldn't encourage CSS or rely on consumers to need CSS overlays. What's the effort in making right-alignment the default?

MartinDoyleUK:
We'd have to clarify the design requirements more carefully. That's an exceedingly specific use case, and we'd need more information on how to flow the text if it was too large, and also what to do with a large quantity of widgets. I implemented what the SFS team requested ... although I'm of course happy to iterate if necessary.

draperd:
It is an exceedingly specific use case, but we only have one consumer of it - and whilst the design is not perfect it makes sense to implement what little design we do have. It might be possible to reuse the alfresco/layout/LeftAndRight module here for a quick layout solution. Until we have another requirement to align the notification actions differently I don't see why we should.

draperd:
Looks good!

PR #981, opened 2016-04-08
AKU-920, AKU-921: Expanded cell form control focus and cursor keys use

draperd:
This PR jointly addresses both https://issues.alfresco.com/jira/browse/AKU-920 and https://issues.alfresco.com/jira/browse/AKU-921 to ensure when form controls are present in the expanded area of a list or grid that they can be given focus and the cursor keys can be used as expected without controlling the item focused in the enclosing list. 

**FULL REGRESSION SUITE RUNNING - REVIEW DON'T MERGE!**

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
@@ -88,6 +88,7 @@
  * @mixes module:alfresco/forms/controls/FormControlValidationMixin
  * @mixes module:alfresco/core/Core
  * @mixes module:alfresco/forms/controls/utilities/RulesEngineMixin
+ * @mixes module:alfresco/lists/KeyboardNavigationSuppressionMixin
  * @extendSafe
  * @author Dave Draper
  * @author Richard Smith
@@ -301,6 +302,22 @@ define(["dojo/_base/declare",
        * @since 1.0.33
        */
       getPubSubOptionsImmediately: true,
+
+      /**
+       * This attribute should only be configured to be true when the control is being used within the
+       * a [list]{@link module:alfresco/lists/AlfList} - for example as part of the widget model for
+       * [a list appendix]{@link module:alfresco/lists/views/AlfListView#widgetsForAppendix} or as 
+       * part of the rendering of an item. Configuring this to true will ensure that using the cursor
+       * keys will not control the selected list item and that focus will not be stolen from the wrapped
+       * control - however, this should be used with caution as it may limit or alter the behaviour
+       * of some form controls. See AKU-920/AKU-921 for the background of why this has been added.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.63
+       */
+      usedInList: false,
 
       /**
        * The default visibility status is always true (this can be overridden by extending controls).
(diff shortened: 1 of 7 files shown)


draperd:
I've pushed some updates following regression test failures and am now re-running the full suite. The second commit can be reviewed though.

aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -303,6 +304,22 @@ define(["dojo/_base/declare",
...
+       * keys will not control the selected list item and that focus will not be stolen from the wrapped
+       * control - however, this should be used with caution as it may limit or alter the behaviour
+       * of some form controls. See AKU-920/AKU-921 for the background of why this has been added.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       * @since 1.0.63
+       */
+      usedInList: false,
```
MartinDoyleUK: This feels horribly like "coding by exception". Is this definitely the best way to fix the problem?

aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -1261,14 +1278,23 @@ define(["dojo/_base/declare",
...
          }
          else
          {
             // No action yet. Don't unsubscribe.
          }
       },
 
+      suppressFocusRequest: function alfresco_lists_KeyboardNavigationSuppressionMixin__suppressFocusRequest(evt) {
+         evt.stopPropagation();
+         evt.preventDefault();
```
MartinDoyleUK: We should try to ensure we only use one of these, generally `stopPropagation()`, otherwise we're adding redundant code. Just best-practice rather than anything wrong with what's present.

aikau/src/test/resources/config/Suites.js
```diff
@@ -27,11 +27,17 @@ define(function() {
...
    // This is the collection to change when only some tests are required
    var someTests = [
-      "src/test/resources/alfresco/upload/UploadMonitorTest"
+      "src/test/resources/alfresco/forms/controls/CheckBoxTest",
+      // "src/test/resources/alfresco/forms/controls/FormButtonDialogTest",
+      // "src/test/resources/alfresco/forms/controls/PushButtonsTest",
+      // "src/test/resources/alfresco/search/SearchSuggestionsTest",
+      // "src/test/resources/alfresco/core/PageTest",
+      // "src/test/resources/alfresco/documentlibrary/views/MediaFilmStripViewTest",
+      // "src/test/resources/alfresco/forms/controls/ContainerPickerTest"
```
MartinDoyleUK: Do we need all the commented-out tests to be added?

aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
draperd: I meant to remove that - it's no longer necessary, I'll get rid of it

aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
draperd: Well, it's **a** solution. The problem here is that the RM use case probably won't be a common one. We need to both fix the issue and not screw up the form controls. It's the best I can come up with at the moment.

MartinDoyleUK: Good reason! :+1:

draperd: Improved solution in #982

PR #993, opened 2016-04-13
AKU-893: Paginator configuration updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-893 to provide additional configuration options for the alfresco/lists/Paginator widget:

- showPageSelector (true/false)
- currentPageLabelStyle ("NUMBER", "X_OF_Y")
- backLabel
- forwardLabel

Unit tests have been updated to verify the new options.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/Paginator.js
+++ b/aikau/src/main/resources/alfresco/lists/Paginator.js
@@ -88,6 +88,16 @@ define(["dojo/_base/declare",
       i18nRequirements: [{i18nFile: "./i18n/Paginator.properties"}],
       
       /**
+       * The label for the page back menu item.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.64
+       */
+      backLabel: "list.paginator.back.label",
+
+      /**
        * Indicates whether the paginator should be displayed in "compact" mode where only
        * the back and forward buttons are displayed.
        *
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/lists/Paginator.js
```diff
@@ -105,6 +115,18 @@ define(["dojo/_base/declare",
...
+       * The style of current page number. The default is "NUMBER" which indicates that just the
+       * page number should be displayed. The only alternative in the current version is
+       * "X_OF_Y" which will render the label in the form "1 of 4" (as in page 1 of 4).
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.64
+       */
+      currentPageLabelStyle: "NUMBER",
```
MartinDoyleUK: Is it worth making this a format mask instead, e.g. `"{pageNum} of {numPages}"` or `"Numero {pageNum}"`? Dunno how much extra effort is involved and whether there would be any problems with i18n. Perhaps it's a supplementary feature?

aikau/src/main/resources/alfresco/lists/Paginator.js
```diff
@@ -468,7 +521,7 @@ define(["dojo/_base/declare",
                }
                
                // Create the page labels, which for English will be along the lines of 1-25
-               if (this.compactMode === false)
+               if (this.compactMode === false && this.showPageSelector)
```
MartinDoyleUK: Should `this.compactMode` be converted into `this.showPageSelector`?

draperd: No, because we had compactMode and can't remove it (backwards compatibility) and showPageSelector does something different.

aikau/src/main/resources/alfresco/lists/Paginator.js
draperd: I thought about that, but for just two options I didn't think it would be worth it. Also, it's SFS specific at the moment

PR #1013, opened 2016-04-21
AKU-937, AKU-933: Added support for view modifiers as used in DocumentListPicker

draperd:
This PR addresses both https://issues.alfresco.com/jira/browse/AKU-933 and https://issues.alfresco.com/jira/browse/AKU-937 to ensure that it is possible to process view models on each rendering of a view. This approach is then used on the DocumentListPicker to ensure that proper cloning of the view model is done before rendering. A unit test has been added to prevent future regressions.

### DIFF
--- a/aikau/src/main/resources/alfresco/pickers/DocumentListPicker.js
+++ b/aikau/src/main/resources/alfresco/pickers/DocumentListPicker.js
@@ -133,8 +133,11 @@ define(["dojo/_base/declare",
        * @instance
        */
       postCreate: function alfresco_pickers_DocumentListPicker__postCreate() {
-         this.processObject(["processInstanceTokens"], this.widgets);
-         this.processWidgets(this.widgets, this.itemsNode);
+         // NOTE: Need to set viewModifiers for all subsequent calls to renderView...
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/pickers/DocumentListPicker.js
```diff
@@ -133,8 +133,11 @@ define(["dojo/_base/declare",
        * @instance
        */
       postCreate: function alfresco_pickers_DocumentListPicker__postCreate() {
-         this.processObject(["processInstanceTokens"], this.widgets);
-         this.processWidgets(this.widgets, this.itemsNode);
+         // NOTE: Need to set viewModifiers for all subsequent calls to renderView...
+         this.viewModifiers =  ["processInstanceTokens"];
```
MartinDoyleUK: Doesn't this override any custom-specified ones in the model? Is that intentional?

draperd: Yes, DocumentList picker is a very specific implementation - without this modifier, it's not going to work.

Issue #999, opened 2016-04-14
Form controls inside a ControlRow in a Dialog created using the ALF_CREATE_DIALOG_REQUEST not publishing to _valueChangeOf_ topic when form is rendered


johhnywishwash:
Form controls, that are inside a **_ControlRow_**, that are in a **_Form_**, that have been created in a Dialog using **ALF_CREATE_DIALOG_REQUEST** will not initially publish their value to the _valueChangeOf_ topic unless the Forms config has the the property **_"value:{}"_**.

I came across this as we are creating a dialog that contains a **_DynamicForm_** that is updated by a select control in the same dialog window.  I noticed that only the controls not inside **_ControlRows_** were publishing to the **_valueChangeOf_** topic.

A **_Form_** without the **_"value:{}"_** set but just on the page seems to work fine.

As a work around I have just included **_"value:{}"_** in the **_Form_** config
Here is an example demonstrating the behaviour.  If you filter the PubSub log by _valueChangeOf_ you will see Control C does not publish its value:

```
{
		name : "alfresco/forms/Form",
		config : {
			widgets : [ {
				name : "alfresco/forms/ControlRow",
				config : {
					title : "Control Row 1",
					widgets : [ {
						name : "alfresco/forms/controls/TextBox",
						config : {
							fieldId : "control_a",
							name : "control_a",
							label : "Control A"
						}
					} ]
				}
			}, {
				name : "alfresco/forms/controls/TextBox",
				config : {
					fieldId : "control_b",
					name : "control_b",
					label : "Control B"
				}
			} ]
		}
	}, {
		name : "alfresco/buttons/AlfButton",
		config : {
			label : "Show Form",
			publishTopic : "ALF_CREATE_DIALOG_REQUEST",
			publishPayload : {
				dialogId : "DIALOG",
				dialogTitle : "DIALOG FORM WITH CONTROL ROW",
				widgetsContent : [ {
					name : "alfresco/forms/Form",
					config : {
						//value : {},
						widgets : [ {
							name : "alfresco/forms/ControlRow",
							config : {
								title : "Control Row 2",
								widgets : [ {
									name : "alfresco/forms/controls/TextBox",
									config : {
										fieldId : "control_c",
										name : "control_c",
										label : "Control C"
									}
								} ]
							}
						}, {
							name : "alfresco/forms/controls/TextBox",
							config : {
								fieldId : "control_d",
								name : "control_d",
								label : "Control D"
							}
						} ]
					}
				} ],
[...]

PR #1014, opened 2016-04-21
AKU-938: Control in control row in dialog with no value in form fix

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-938 / #999 to ensure that ControlRow widgets behave in the same way regardless of whether or not they are placed in a dialog and whether or not the form they are in are assigned a value. This was quite a tricky problem to solve and a number of different options were explored. In the end it was determined that the best approach was always to assign the form a value (even if empty). This resulted in some other changes being necessary to form controls and existing tests that were written in a sub-optimal way. Also, it was found that it was necessary to rework the PubQueue singleton to ensure that all publications were truly made in the intended order.

**DO NOT MERGE YET !!! - FULL REGRESSION TEST RUN IN PROGRESS**

### DIFF
--- a/aikau/src/main/resources/alfresco/core/PubQueue.js
+++ b/aikau/src/main/resources/alfresco/core/PubQueue.js
@@ -50,6 +50,19 @@ define(["dojo/_base/declare",
       _released: false,
 
       /**
+       * To make sure that all publications are made in order after releasing, this flag is used to indicate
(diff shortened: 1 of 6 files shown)


draperd:
The proposed fix is causing problems with both the drag-and-drop framework and form controls that accept multiple values from a set of options

PR #1021, opened 2016-04-27
AKU-951: Ensure that validation message is displayed when a form control has no label

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-951 to ensure that form controls will always display error messages even if they have no label. The unit tests have been updated to ensure this and an additional test has been added to ensure that error messages are shown in form dialogs (because currently the form dialogs have different layout)

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
@@ -1249,7 +1249,6 @@ define(["dojo/_base/declare",
             }
             else
             {
-               // domStyle.set(this._titleRowNode, {display: "none"});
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -1245,7 +1249,8 @@ define(["dojo/_base/declare",
             }
             else
             {
-               domStyle.set(this._titleRowNode, {display: "none"});
+               // domStyle.set(this._titleRowNode, {display: "none"});
```
MartinDoyleUK: Shouldn't this line just be deleted?

draperd: Yes. Yes, it should.

PR #1025, opened 2016-04-28
AKU-953: SelectedItemStateMixin fix

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-953 to ensure that the SelectedItemStateMixin treats all values appropriately. A new unit test has been added to verify and prevent regressions.

### DIFF
--- a/aikau/src/test/resources/alfresco/renderers/SelectorTest.js
+++ b/aikau/src/test/resources/alfresco/renderers/SelectorTest.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright (C) 2005-2016 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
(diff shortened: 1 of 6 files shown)


aikau/src/test/resources/alfresco/renderers/SelectorTest.js
```diff
@@ -0,0 +1,72 @@
...
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * This test renders examples of FileTypes.
+ *
+ * The test is simple and much of its validity is in the use of slightly damaged or incomplete models to inspect edge cases.
+ *
+ * @author Richard Smith
+ */
```
MartinDoyleUK: This comment block seems wrong??

aikau/src/test/resources/config/Suites.js
```diff
@@ -252,6 +252,7 @@ define(function() {
       "alfresco/renderers/PublishPayloadMixinOnActionsTest",
       "alfresco/renderers/ReorderTest",
       "alfresco/renderers/SearchResultPropertyLinkTest",
+      "src/test/resources/alfresco/renderers/SelectorTest",
```
MartinDoyleUK: I didn't explicitly mention it (apologies) but I discovered that we don't need the src/test prefix, and so it was removed from all tests.

aikau/src/test/resources/testApp/WEB-INF/classes/alfresco/site-webscripts/alfresco/renderers/Selector.get.desc.xml
```diff
@@ -0,0 +1,7 @@
+<webscript>
+  <shortname>Selector Renderer</shortname>
+  <description>Displays Selector renderers for testing.</description>
+  <description></description>
```
MartinDoyleUK: Duplicate element

PR #1029, opened 2016-04-29
AKU-941: Update thumbnail to support aspect specific folder rendering

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-941 to make it possible to configure a mapping for aspects on a folder node to the image that is used to render that folder. In particular this adds default support for Smart Folders. The unit tests have been updated.

**FULL REGRESSION RUN IN PROGRESS- REVIEW BUT DON'T MERGE**


aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -522,6 +548,14 @@ define(["dojo/_base/declare",
          this.imgAltText = "";
          this.imgTitle = "";
 
+         // See AKU-941 - support for smart folders
+         if (!this.folderImageAspectMappings)
```
MartinDoyleUK: Should this be additive, if a mapping definition already exists?

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
```diff
@@ -611,7 +645,24 @@ define(["dojo/_base/declare",
        * @instance
        */
       getFolderImage: function alfresco_renderers_Thumbnail__getDefaultFolderImage() {
-         return require.toUrl("alfresco/renderers") + "/css/images/" + this.folderImage;
+         var url, jsNode = this.currentItem.jsNode;
+         if(jsNode && jsNode.aspects)
+         {
+            for (var key in this.folderImageAspectMappings)
+            {
+               if(this.folderImageAspectMappings.hasOwnProperty(key) && jsNode.aspects.indexOf("smf:smartFolder") !== -1)
```
MartinDoyleUK: This test ... `jsNode.aspects.indexOf("smf:smartFolder") !== -1` only needs to happen once, not on every loop

aikau/src/main/resources/alfresco/renderers/Thumbnail.js
draperd: No, I don't think so at this stage - because then we'd have to support clever ways of removing existing mappings, etc.

PR #1034, opened 2016-05-03
Eager-resolve preferences loaded when provided via config

AFaust:
This minor change ensures that localPreferences - if set via configuration like share-header.lib.js does - will be used as the current preferences and early-resolve the promise. This ensures that e.g. the LoggingService can retrieve the logging preferences of the user without waiting for an asynch getPreference call or a log menu publishing changes (which may not be included in the page), which means that LoggingService will correctly log messages from the beginning instead of potentially missing some initial page load log requests.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/PreferenceService.js
+++ b/aikau/src/main/resources/alfresco/services/PreferenceService.js
@@ -86,13 +86,21 @@ define(["dojo/_base/declare",
 
          // Load the user preferences...
          this._preferencesLoaded = new Deferred();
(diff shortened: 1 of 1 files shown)


draperd:
Thanks for raising this - I'm going to check out the branch locally, run the unit tests and test against Share and if no problems show up then I'll merge later today for inclusion in the 1.0.67 release (which is going to happen this Friday, rather than next Tuesday due to team vacation time)

draperd:
I've not found any issues so am merging - many thanks for the contribution - hopefully the first of many! :)

PR #1040, opened 2016-05-05
AKU-957: Ensure CommentsList refreshes on comment edit

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-957 to ensure that CommentsList refreshes when a comment is edited inline.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/EditableComment.js
+++ b/aikau/src/main/resources/alfresco/renderers/EditableComment.js
@@ -236,7 +236,7 @@ define(["dojo/_base/declare",
        * @instance
        * @param {object} payload The details of the comment update
        */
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/renderers/EditableComment.js
```diff
@@ -236,7 +236,7 @@ define(["dojo/_base/declare",
        * @instance
        * @param {object} payload The details of the comment update
        */
-      onEditSave: function alfresco_renderers_EditableComment__onEditSave(payload) {
+   onEditSave: function alfresco_renderers_EditableComment__onEditSave(payload) {
```
MartinDoyleUK: This looks like an outdent that then mis-aligns the function definition ... is that right?

PR #1042, opened 2016-05-06
AKU-942: Extend drop target in AlfDocumentList to entire available space

MartinDoyleUK:
This addresses issue [AKU-942](https://issues.alfresco.com/jira/browse/AKU-942) to allow display of a message for empty document lists (in the context of being able to upload to it and not), and also to allow a minimum height to be set (with a default for document lists) for all lists.


aikau/src/main/resources/alfresco/lists/views/EmptyDocumentList.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+
+/**
+ * This widget serves to display information and functionality relevant for displaying in an empty document list.
+ *
+ * @module alfresco/lists/views/EmptyDocumentList
+ * @extends external:dijit/_WidgetBase
+ * @mixes external:dojo/_TemplatedMixin
+ * @mixes module:alfresco/core/Core
+ * @author Martin Doyle
```
draperd: Since ?

PR #1045, opened 2016-05-07
Avoid double pubSubScope-prefix in logged publications

AFaust:
This minor change fixes a publication logging issue with pubSubScope / alfPublishScope being added as a redundant prefix to the already scoped alfTopic.


draperd:
Thanks for spotting and fixing this!

PR #1046, opened 2016-05-07
Add missing message call to iconAltText case

AFaust:
This minor change ensures that iconAltText is properly set (looking up in i18n resources) in one specific case of the _AlfMenuItemMixin postMixInProperties.

### DIFF
--- a/aikau/src/main/resources/alfresco/menus/_AlfMenuItemMixin.js
+++ b/aikau/src/main/resources/alfresco/menus/_AlfMenuItemMixin.js
@@ -164,7 +164,7 @@ define(["dojo/_base/declare",
          }
          if (!this.iconAltText && this.title)
          {
-            this.iconAltText = originalLabel;
+            this.iconAltText = this.message(originalLabel);
          }
          else if (this.iconAltText)
          {


aikau/src/main/resources/alfresco/menus/_AlfMenuItemMixin.js
```diff
@@ -164,7 +164,7 @@ define(["dojo/_base/declare",
          }
          if (!this.iconAltText && this.title)
          {
-            this.iconAltText = originalLabel;
+            this.iconAltText = this.message(originalLabel);
```
draperd: Looking at the original code, I'm wondering if my intention (and the more appropriate behaviour) would be to use `this.title` for the `iconAltText`... because it's likely that the title would be more informative than the label. I'm happy to merge this (as it's a fairly minor change I'm not worried about a unit test), but just wanted to flag up that it might be better to use the title rather than the label).

AFaust: I think it is a pure semantic argument wether label or title should be the go-to fallback value for altText if not configured. I'd expect the label to be used for altText of the icon whereas title would be expected to be used in the DOM title attributes of both the label and the icon for the menu item.

draperd: OK, I think I'll just merge as is then.

Issue #1049, opened 2016-05-10
Incorrect _valueChangeOf_ payload if the publication is called by autoSet


aviriel:
It's me again. I tried to find a hack for #1047 and decided to create new hidden control on the form, the value of which will be updated automatically depending on values of two other controls. So I can make a visibilityConfig for a label on the form depending on this one control. 

VisibilityConfig for a label "works" if I set the value of PRESET manually (it's a textbox), but if I change values of Select fields, then value of PRESET is updating, but label visibility is not changing. Looking at PubSub log, I can say that if the value is set by autoSetConfig, then _valueChangeOf payload does not contain "value" field. How can it happen and what is the way to solve it?

Automatically set value:
![automatically set value](https://cloud.githubusercontent.com/assets/596948/15145715/e8b8d214-16bf-11e6-88a1-c523520e16c2.png)

Manually set value:
![manually set value](https://cloud.githubusercontent.com/assets/596948/15145718/ebe267f2-16bf-11e6-9202-756dc9e221b3.png)

Form:
```
    {
    	name: "alfresco/forms/DynamicForm",
    	config: {
    		publishValueSubscriptions: ["ALF_ENTER_KEY_PRESSED"],
    		okButtonPublishTopic: "TEST",
    		showCancelButton: false,
    		widgets: [{
    			name: "alfresco/forms/controls/Select",
    			config: {
    				name: "A",
    				label: "A",
    				fieldId: "PRESET_A",
    				value: "1",
    				optionsConfig: {
    					fixed: [{
    						"label": "1",
    						"value": "1"
    					}, {
    						"label": "2",
    						"value": "2"
    					}]
    				}
    			}
    		}, {
    			name: "alfresco/forms/controls/Select",
    			config: {
    				name: "B",
    				label: "B",
    				fieldId: "PRESET_B",
    				value: "1",
    				optionsConfig: {
    					fixed: [{
    						"label": "1",
    						"value": "1"
    					}, {
    						"label": "2",
    						"value": "2"
    					}]
    				}
    			}
    		}, {
    			name: "alfresco/forms/controls/TextBox",
    			config: {
    				name: "hidden",
    				fieldId: "PRESET",
    				value: "11",
    				autoSetConfig: [{
    					rulePassValue: "11",
    					rulesMethod: "ALL",
    					rules: [{
    						targetId: "PRESET_A",
    						is: ["1"]
    					}, {
    						targetId: "PRESET_B",
    						is: ["1"]
    					}]
[...]

PR #1051, opened 2016-05-11
#1049 - only set rulePassValue / ruleFailValue if actual value is defined

AFaust:
After checking @aviriel s issue #1049 I realized this is fixed by one small change to autoSetValue that we had already implemented as one of our 160+ fixes/changes.
This change avoids setting the value of a widget to undefined and thus allows rulePassValue or ruleFailValue to not be defined, so a developer can define rules that only change the value on positive or negative rule match, leaving the previous value intact otherwise. In the case of @aviriel the previous value may be a value of a different autoSetConfig rule set, since each element of the autoSetConfig is individually evaluated and her example triggers all the rules.


draperd:
There is no unit test with this PR so I've raised https://issues.alfresco.com/jira/browse/AKU-970 to create the necessary unit test to accompany this proposed fix for #1049.

draperd:
This has now been replaced by #1070

Issue #1050, opened 2016-05-11
When rendering date with alfresco/renderers/Date and simple: true the label isn't internationalised


toonconvents:
When using the alfresco/renderers/Date to render a date with simple: true (to render the supplied date value instead of the modified and created) the label isn't replaced by its property value.
Example:
{
    name: "alfresco/renderers/Date",
    config: {
        simple: true,
        propertyToRender: "someDateProperty",
        renderedValueClass: "alfresco-renderers-Property pointer",
        renderSize: "small",
        label: "some.label.key.in.the.properties.file"
    }
}

draperd:
Apologies for the delay in responding to this issue - the Aikau team were on vacation, we will start looking through the backlog of issues from today - thanks for your patience!

draperd:
This looks to be a bug and as such we would prioritize for the next sprint... I can see that @AFaust has already raised a PR to address this issue, but we like to include unit tests to prevent future regressions. I'll raise a JIRA ticket for this issue that will include the change that @AFaust has provided but to augment with a unit test.

draperd:
I've raised https://issues.alfresco.com/jira/browse/AKU-969 to cover this issue - as a bug it will be prioritized by our next sprint, starting next Tuesday. The fix can be expected in the 1.0.69 release.

draperd:
The fix for this is now available in version 1.0.68

PR #1055, opened 2016-05-12
#1050 - add label handling and align with Property.js render logic

AFaust:
This small change addresses the missing label i18n handling described in #1050 and applies the current render logic pattern of Property.js for slightly more consistency.


AFaust:
Closing and creating a new PR (targeted the wrong branch).

PR #1056, opened 2016-05-12
#1050 - add label handling and align with Property.js render logic

AFaust:
This small change addresses the missing label i18n handling described in #1050 and applies the current render logic pattern of Property.js for slightly more consistency.


draperd:
Thanks for creating the fix for this - as it is a bug I've raised a JIRA ticket (https://issues.alfresco.com/jira/browse/AKU-969) to ensure that we add the necessary unit test to prevent future regressions. As a bug it will be included in the next sprint, we will review and merge your code to ensure the issue is fixed.

draperd:
Closing as now replaced by #1065 on which this PR was based.

PR #1058, opened 2016-05-16
Update NPM versions

MartinDoyleUK:
Update NPM versions to allow LESS to work with Node v6 and also optimise some code in ConcurrentReporter.


davidcognite:
I've not tested this combination of versions, but a reliable source confirms that they work.

Issue #1059, opened 2016-05-16
alfresco/core/ObjectProcessingMixin processObject function converts array to string when it is inside an object


johhnywishwash:
I have come across an issue where the **_processObject_** function in **_alfresco/core/ObjectProcessingMixin_** will convert a property with an array as its value into a string.  It seems that when processing an object against the currentItem the **_processObject_** function will take the value in the currentItem: 

```
{
    property : ["A","B","C"]
}
```

and replace the token with the array.
If the currentItem is structured so the property is inside in an object

```
{
    object : {
             property : ["A","B","C"]
    }
}
```

then it converts the property into a string and replaces the token with a string.

I have created an example of this issue against version 1.0.67.  I have created an example similar to the **_ObjectProcessingMixinButton_** test.

The page config is : 

```
model.jsonModel = {
	widgets : [ {
		id : "TEST_REPLACE_TOKEN_WITH_ARRAY_IN_OBJECT",
		name : "test/ObjectProcessingMixinButton",
		config : {
			label : "foo.bar",
			token : "{foo.bar}"
		}
	}, {
		id : "TEST_REPLACE_TOKEN_WITH_ARRAY",
		name : "test/ObjectProcessingMixinButton",
		config : {
			label : "bar",
			token : "{bar}"
		}
	}, {
		name : "alfresco/logging/DebugLog"
	} ],
	services : [ ]
};
```

And the test button widget code : 

```
define([ "alfresco/buttons/AlfButton", "alfresco/core/ObjectProcessingMixin",
		"dojo/_base/declare" ], function(AlfButton, ObjectProcessingMixin,
		declare) {

	return declare([ AlfButton, ObjectProcessingMixin ], {

		token : "",

		currentItem : null,

		label : "TEST",

		postMixInProperties : function() {
			this.inherited(arguments);
		},

		onClick : function alfresco_buttons_AlfButton__onClick(evt) {

			this.currentItem = {
				foo : {
					bar : [ "A", "B", "C" ]
				},
				bar : [ "A", "B", "C" ]
			};

			// Publish generated payload
			var payload = this.processObject([ "processCurrentItemTokens" ],
					this._createObject());
			this.alfPublish("GENERATED", payload, true);

			// Stop the event
			if (evt) {
				evt.stopPropagation();
			}
			
		},

		_createObject : function() {
			var obj1 = {
				foo : {
					bar : this.token
				}
			};
			return obj1;
		}
	});
});
```

[...]

PR #1061, opened 2016-05-18
Use lang.getObject() in processTokens and add two standard functions

AFaust:
This PR addresses the issue #1059 by using lang.getObject() to retrieve values from non-trivial object structures without inherently converting them to string (as long as token makes up the entire expression).

This PR also adds two standard processor functions that we have found useful in our internal project(s).
**processHashTokens** allows any generated/processed payload to easily include data from the current hash without requiring the complexities that AlfDynamicPayloadButton uses to mixin/map hash data.
**processMessageTokens** allows any generated/processed payload to include localised strings without forcing the developer to perform "early-localisation", i.e. via server-side msg() calls when constructing a payload or client-side message() calls before alfPublish. This function is not intended to absolve widgets from properly performing localisation of any label / title / description configuration parameters before displaying them, but may be used to bridge gaps (when a 3rd-party module fails to properly localise) or hook in other, locale-sensitive parameters (we use it to load locale-specific custom date/time patterns).

### DIFF
--- a/aikau/src/main/resources/alfresco/core/ObjectProcessingMixin.js
+++ b/aikau/src/main/resources/alfresco/core/ObjectProcessingMixin.js
@@ -29,8 +29,9 @@
 define(["dojo/_base/declare",
         "dojo/_base/lang",
         "dojo/_base/array",
-        "alfresco/core/ObjectTypeUtils"],
-        function(declare, lang, array, ObjectTypeUtils) {
+        "alfresco/core/ObjectTypeUtils",
+        "alfresco/util/hashUtils"],
+        function(declare, lang, array, ObjectTypeUtils, hashUtils) {
 
    return declare(null, {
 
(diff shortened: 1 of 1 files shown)


draperd:
Hi Axel, I can't see anything immediately wrong with what you wish to add, the problem is that you're not providing a unit test to verify the behaviour. I know that you've reported that you've had issues getting the unit test environment up and running, but it is vitally important that we have unit tests to support code changes.

You should think of this as being in your interests - if you provide a unit test then it means that the code you're adding will not be regressed between releases. 

If you're unable to get the unit test running or don't wish to create a unit test then we can create a feature request and place it on the backlog for us to implement at a later date, however we won't merge this pull request until we start work on that feature.

draperd:
I've raised https://issues.alfresco.com/jira/browse/AKU-971 to cover #1059 and to add the necessary unit tests to verify both the bug and the fix (as well as to test the additional helper functions you've provided).

MartinDoyleUK:
Replaced by #1069.

PR #1062, opened 2016-05-20
AKU-956: Updated for handling TinyMCE change events

draperd:
This PR addresses both https://issues.alfresco.com/jira/browse/AKU-956 and https://issues.alfresco.com/jira/browse/AKU-961 (which is a duplicate) to ensure that the TinyMCE editor issues change events on key up to ensure that form buttons can be enabled/disabled on content change rather than a shift in focus. Additionally the edit as well as create comment requires data. Also, when cancelling editing a comment the original value is reset. Unit tests have been updated, unfortunately I've had to reintroduce sleep calls for reliability of testers.

### DIFF
--- a/aikau/src/test/resources/alfresco/renderers/CommentsListTest.js
+++ b/aikau/src/test/resources/alfresco/renderers/CommentsListTest.js
@@ -119,6 +119,11 @@ define(["module",
             });
       },
 
+      "Remove comment content to check save is disabled": function() {
+         return this.remote.execute("tinymce.get(1).setContent('');")
+            .findByCssSelector(".alfresco-renderers-EditableComment .alfresco-forms-Form .confirmationButton");
+      },
+
       "Save edit": function() {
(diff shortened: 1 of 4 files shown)


aikau/src/test/resources/alfresco/renderers/CommentsListTest.js
```diff
@@ -175,50 +180,64 @@ define(["module",
             .click()
          .end()
 
-         .waitForDeletedByCssSelector(".alfresco-lists-AlfList--loading")
-         .end()
+         .getLastPublish("COMMENTS1_ALF_DOCLIST_SORT")
+         .clearLog()
+
+         .sleep(2000)
```
MartinDoyleUK: :(

draperd: I know, I know.... I tried but couldn't reliably get it working :(

PR #1070, opened 2016-05-24
AKU-970: Incorrect _valueChangeOf_ payload if the publication is called by autoSet

draperd:
This PR builds on #1051  to fix https://issues.alfresco.com/jira/browse/AKU-970 ( #1049 ). The original fix provided by @AFaust  has been retained but additional unit tests have been added to verify the fix. I've also taken the opportunity to update a unit test to adopt the best practices of using the DebugLog and reuse the existing test selector properties.


aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -541,13 +541,20 @@ define(["dojo/_base/declare",
        * @param {boolean} hasPassedRule Indicated whether or not evaluation of the rules were successful
        */
       autoSetValue: function alfresco_forms_controls_BaseFormControl__autoSetValue(passValue, failValue, hasPassedRule) {
+         // passValue/failValue may not be set to any value in autoSetConfig to achieve positive-/negative-only rule behaviour
          if (hasPassedRule === true)
          {
-            this.setValue(passValue);
+            if (passValue !== undefined)
```
MartinDoyleUK: Use `typeof === "undefined"` instead, as `undefined` could be overridden by an external library.

_Best practice coding, rather than any great likelihood of it happening_

aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -541,13 +541,20 @@ define(["dojo/_base/declare",
...
+            {
+                this.setValue(passValue);
+            }
          }
          else
          {
-            this.setValue(failValue);
+            if (failValue !== undefined)
```
MartinDoyleUK: As above

Issue #958, opened 2016-03-28
Rendering of large widget structures / AlfList views can take a very long time


AFaust:
The current way that complex, nested widget structures are rendered includes various inefficiencies that appear to have a nasty, non-linear effect on the time it takes to render e.g. a list of 50 entries with 50 very simple columns.

The demonstrator test case in my gist https://gist.github.com/AFaust/1d1450ffe8829371ad04 renders such a list of 50 entries with 50 columns. On IE 11 it can take around 30 seconds for the initial rendition. Firefox (45) appears to be better at handling the use case with around 6-10s. Since I don't use (or even install) Chrome, I can't say how it fares.
In other use cases when we have used a paginating, sortable list, we've additionally observed that destroying the previous page can also take a significant time (~5s on IE11 for a 50 rows and ~20 coumns case).

Sure, any use case that involves excessive complexity can expect to suffer some performance degredation. But it is hard to explain to an end-user / customer, why only 50x50 values are enough to grind his web app to a halt.

In the last week and a half I have built completely separate "high performance" variants of the ListRenderer, Row, Cell, Property and InlineEditProperty widgets (and mixins CoreWidgetProcessing / MultiItemRendererMixin), and also applied some optimisations to secondary widgets such as the Actions renderer, to effectively bring down rendering time to ~10-15% and nearly eliminate overhead of destroying the previous list page.

The main issues we identified as causing significant overheads were:
- construction of the entire view in the live DOM (i.e. each widget is attached to the live DOM during dijit/_WidgetBase create() and only then are children processed)
- use of replacement tokens in widget HTML templates disables caching in dijit/_TemplatedMixin
- synchronous destruction of the entire view while still in the live DOM
- relatively expansive workaround in Dojo 1.10 dom-construct _empty() for potential memory leak in IE <= 9 (https://github.com/dojo/dojo/blob/master/dom-construct.js#L304)
- generic processing of widget template for potential attach points / events (scales with number of tags in HTML)
- premature initialization of secondary widgets / DOM structures that may not be used on every element (Actions menu, inline edit DOM fragments, which could be initialized on the first "mouse hover" of a duration of 100ms or more)
- application of inline styles (declared in widget config) on each instance instead of using dynamically applied CSS rules, i.e. using a shared cell/column/property identifier
- evaluation / application of ALL declarative widget config parameters as potential DOM attributes in dijit/_WidgetBase _applyAttributes even though 99+% are typically irrelevant
[...]

PR #1083, opened 2016-05-31
Aikau 2 Prototyping

draperd:
**PLEASE DO NOT MERGE THIS PR - FOR COLLABORATION PURPOSES ONLY**

This PR has been opened to show some prototyping work in progress for code that may potentially become the next major version of Aikau. It is intended to be complementary to the existing Aikau code and exists in a standalone package.

The idea is to provide a package for experimental code that will not be backwards compatible until the time comes to make it the primary package for a future release (and deprecate the existing code). This will give us the opportunity to get feedback from the community over multiple releases without the restrictions of needing to support previously releases code.

The primary objectives of this PR are to look at simplifying and improving widget processing through the use of promises and to improve performance by not building directly into the live DOM (as well as other ideas presented in #958 by @AFaust). 

Rather than attempting to retro fit the existing widgets this will look at building widgets using the Material Design (initially using https://getmdl.io/ but may also look at other implementations such as http://materializecss.com/)

This code will be aimed at evergreen browsers and look to make use of modern JS/CSS and not strive for compatibility with earlier versions of Internet Explorer.

Other things to consider will be the use of a virtual DOM (potentially http://maquettejs.org/) and integration of other tools and libraries (although not necessarily under this same PR).

**PLEASE NOTE - THERE IS NO GUARANTEE OR, OR TIMELINE FOR MERGING THIS CODE**


jgoldhammer:
Any progress here?

draperd:
Hi Jens, unfortunately I haven't been able to spend any time working on this since I closed the PR on June 1st... I do intend to pick it up as soon as I have some time to focus on it - but I can't say when that will be. Thanks for asking about it though! It's always good to know that people are interested in this!

PR #1101, opened 2016-06-09
AKU-983: Create generic "processing" indicator

MartinDoyleUK:
This addresses issue [AKU-983](https://issues.alfresco.com/jira/browse/AKU-983) and creates a new feature in `alfresco/services/NotificationService` that can display/hide an "indeterminate" activity indicator. There are three new events in `alfresco/core/topics` that are used to drive the display of the indicator: `PROGRESS_INDICATOR_ADD_ACTIVITY`, `PROGRESS_INDICATOR_REMOVE_ACTIVITY` and `PROGRESS_INDICATOR_REMOVE_ALL_ACTIVITIES`.

The functionality has been added to the `alfresco/services/ManageAspectsService` and `alfresco/services/NavigationService` as well.

A full test suite has been run, relevant regressions fixed, and new tests have been added.


draperd:
Couple of things to investigate and correct before this can be merged...

1) If you test this out on the NavigationService unit test page and click the "Post to New Page" button you'll see the progress indicator - but it never disappears because the link has been opened in a new tab. This needs to be corrected

2) I'm wondering why I'm not seeing the progress indicator when navigating around the unit test application - I thought I would still see it? Maybe everything is moving to fast?

I did test this out in Share though and it does look good and work well when navigating using Aikau controls.

draperd:
I've made the minor change myself as @MartinDoyleUK is not available. Merging

MartinDoyleUK:
😀

PR #1102, opened 2016-06-10
AKU-991: Added node.nodeRef fallback to NodePreviewService

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-991 to provide a fallback option for locating a NodeRef of the Node to preview when it is not explicitly declared in the payload. I've not included a unit test change for this issue as it is a minor additive change and have verified the existing unit tests continue to work.


davidcognite:
LGTM.

PR #1113, opened 2016-06-15
AKU-982 - Add support for disabling click actions in Indicators

skopf:
Simply added a boolean flag to torn off actions on indicators. Default value is current behaviour.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/Indicators.js
+++ b/aikau/src/main/resources/alfresco/renderers/Indicators.js
@@ -81,6 +81,15 @@ define(["dojo/_base/declare",
       propertyToRender: null,
 
       /**
+       * Whether this indicator is clickable and publishes a action request or not.
+       *
+       * @instance
+       * @type {boolean}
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/renderers/Indicators.js
```diff
@@ -81,6 +81,15 @@ define(["dojo/_base/declare",
       propertyToRender: null,
 
       /**
+       * Whether this indicator is clickable and publishes a action request or not.
+       *
+       * @instance
+       * @type {boolean}
+       * @default
+       */
```
draperd: When we add a new configuration attribute or function we always add the @since tag with the version it was added in

draperd:
This is missing a unit test. This issue was already in the sprint plan and would have been in the next release anyway.  I'm going to check out your branch, close this PR, add the test and then create a new PR.

draperd:
This has now been superceded by #1114

PR #1114, opened 2016-06-16
AKU-982: Add support for disabling click actions in Indicators

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-982 and replaces #1113 (which was missing a unit test). It adds a configuration option to the Indicators renderer that will prevent any configured actions for the indicator being supported. A unit test has been added to verify this change.


davidcognite:
LGTM.

PR #1115, opened 2016-06-16
AKU-947: PushButton styling updates

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-947 to modify the PushButton styling. A new configuration option called "simpleLayout" has been provided to enable this new styling. As this is a purely cosmetic change the unit tests have not been updated (although the test page has). Existing unit tests still pass.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/css/PushButtonsControl.css
+++ b/aikau/src/main/resources/alfresco/forms/controls/css/PushButtonsControl.css
@@ -11,12 +11,6 @@
       position: absolute;
       width: 0;
 
-      + label {
-         background: @push-button-background;
-         border-color: @push-button-border-color;
-         color: @push-button-text-color;
-      }
-
       &:checked + label {
          background: @push-button-background-checked;
          border-color: @push-button-border-color-checked;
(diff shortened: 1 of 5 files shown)


aikau/src/main/resources/alfresco/forms/controls/css/PushButtonsControl.css
```diff
@@ -4,38 +4,70 @@
...
+            margin-bottom: @push-button-margin;
+         }
+      }
+   }
+
    &__label {
-      border: 1px solid @button-color-default;
-      border-radius: 10px;
+      border: 1px solid #aaa;
+      border-radius: 3px;
```
davidcognite: Why not use something like @standard-button-border-radius?

aikau/src/main/resources/alfresco/forms/controls/css/PushButtonsControl.css
```diff
@@ -4,38 +4,70 @@
...
+            margin-right: @push-button-margin;
+            margin-bottom: @push-button-margin;
+         }
+      }
+   }
+
    &__label {
-      border: 1px solid @button-color-default;
-      border-radius: 10px;
+      border: 1px solid #aaa;
```
davidcognite: Should this colour really be hard coded rather than themeable?

aikau/src/main/resources/alfresco/forms/controls/css/PushButtonsControl.css
draperd: Good point - but that variable name has already been taken. The push button styling is non-standard, but I will add the missing variables.

aikau/src/main/resources/alfresco/forms/controls/css/PushButtonsControl.css
draperd: You're quite right - actually, this was left over from an earlier attempt - I think it should be removed.

Issue #1104, opened 2016-06-10
Tutorial needs fix for Claring Dependency Caches


douglascrp:
I'm executing the tutorial, using the latest Aikau release available today, which is 1.0.70.
When I tried to follow the instructions from https://github.com/Alfresco/Aikau/blob/master/tutorial/chapters/ClearingDependencyCaches.md I noticed that the Clear Dependencies Cache does not exist anymore.

I'm not sure, but I think the tutorial needs some updates.
Is this because of the Surf dependencies upgrade?
Thank you in advance.

draperd:
Thanks for raising this issue... the reason is that we've moved to Surf 6 to take advantage of performance  improvements. As a result of the Alfresco Share / Repository separation the "Clear Dependency Caches" button has been relocated - it can now be found at the following location: http://localhost:8089/aikau/service/console

I'll make sure to get the tutorial updated as soon as I can.

draperd:
I've raised https://issues.alfresco.com/jira/browse/AKU-993 to address this - thanks again for raising!

douglascrp:
@draperd Ok, I've just tried it and it works.
Thank you.

draperd:
This is now fixed

PR #1116, opened 2016-06-16
AKU-993: Update clear caches tutorial page

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-993 / #1104 to correct the URL and screenshot for clearing dependency caches. This change is required since moving up to Surf 6 which has relocated the clear caches button to a different page.


davidcognite:
Is it worth deleting the old screenshot as part of this commit?

PR #1132, opened 2016-06-28
AKU-1006: Add support for group image in AvatarThumbnail

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1006 to provide support in the alfresco/renderers/AvatarThumbnail for configuring a property and values that determine whether or not the currentItem is a group so that a more appropriate image can be displayed. The unit test has been updated to verify the change.


aikau/src/main/resources/alfresco/renderers/AvatarThumbnail.js
```diff
@@ -112,7 +162,20 @@ define(["dojo/_base/declare",
             this.currentItem[this.userNameProperty] = encodeURIComponent(this.currentItem[this.userNameProperty]);
             this.thumbnailUrlTemplate = "slingshot/profile/avatar/{" + this.userNameProperty + "}/thumbnail/avatar";
          }
+         
          this.inherited(arguments);
+
+         if (this.groupProperty && this.groupValues)
```
MartinDoyleUK: Worth adding `&& this.groupValues.length` to avoid running the condition if there's no values? (Not entirely sure of the logic here)

aikau/src/main/resources/alfresco/renderers/AvatarThumbnail.js
```diff
@@ -112,7 +162,20 @@ define(["dojo/_base/declare",
...
+         {
+            var groupPropertyValue = lang.getObject(this.groupProperty, false, this.currentItem);
+            this.isGroup = array.some(this.groupValues, function(value) {
+               return value === groupPropertyValue;
+            });
+         }
+         if (this.isGroup)
+         {
+            this.thumbnailUrl = require.toUrl("alfresco/renderers/css/images/" + this.groupImage);
```
MartinDoyleUK: `urlUtils` has been used elsewhere for these kind of URLs ... might be good to be consistent?

```javascript
var imagePath = "alfresco/renderers/css/images/" + this.groupImage;
urlUtils.convertUrl(imagePath, urlTypes.REQUIRE_PATH);
```

draperd: The current code is consistent with the how the folder images are handled in Thumbnail so I'm happy to leave this as is.

aikau/src/main/resources/alfresco/renderers/AvatarThumbnail.js
draperd: Pushed an update

PR #1137, opened 2016-07-01
AKU-1010: Ensure that pending list requests are handled

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1010 to ensure that it is possible for AlfList to make an attempt to cancel an in-flight request (although this relies on the service supporting it) and for pending requests to be issued.

In particular this solves the issue of a breadcrumb trail being used when a document list is loading data. This is what the unit test has been set up to verify (using TDD I proved the unit test failed before fixing)

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfList.js
@@ -174,6 +174,18 @@ define(["dojo/_base/declare",
       currentData: null,
 
       /**
+       * An internally used attribute to hold a UUID for the any in-flight request. This allows the request to be
+       * cancelled if a subsequent request is issued before the in-flight request completes. This should not be
+       * configured.
+       * 
+       * @instance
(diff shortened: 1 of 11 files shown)


MartinDoyleUK:
All looks fine to me. I'm happy for you to merge when the regression run has completed.

aikau/src/main/resources/alfresco/lists/AlfList.js
```diff
@@ -350,6 +362,16 @@ define(["dojo/_base/declare",
       requestInProgress: false,
 
       /**
+       * An internally used attribute used to indicate whether or not a request is pending to be performed
+       * as soon as the current request completes.
+       * 
+       * @instance
+       * @type {boolean}
+       * @default
+       */
```
draperd: Note - there's no @since statement here - but this is because it was previously present, but just not declared as an attribute.

PR #1142, opened 2016-07-07
AKU-1017: Prevent notifications being shown on download related actions

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1017 to ensure that the progress notification isn't shown when using actions such as download or edit offline that use a special request parameter on the URL to trigger a download. A unit test has been added to verify the fix and prevent regressions.


davidcognite:
LGTM

PR #1143, opened 2016-07-07
AKU-1018: Specific Firefox handling of click in PublishAction

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1018. I'm not sure why the original code was failing in Firefox (and only Firefox) but this does seem to fix the problem.


davidcognite:
LGTM.

PR #1145, opened 2016-07-07
AKU-1015: Forms runtime support initial proof of concept

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1015 to provide an initial proof of concept for rendering XML forms configuration in Share using Aikau. It adds a new project to provide a new JAR file containing a new version of the Java-backed WebScript for the Forms Runtime that returns JSON data (rather than an HTML fragment). There is also a new FormsService that can request form data from this WebScript and provide the Aikau form.

The output does not yet include mappings for all Share form controls or all constraints, but provides enough basic features to gather feedback from the community for bug fixes and enhancements.


MartinDoyleUK:
Can't see any obvious problems anywhere.

PR #1148, opened 2016-07-11
AKU-1020: Focus items on list load

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1020 to provide support for item focus in lists (to typically address the use case of selecting an item that has just been created). A unit test has been updated to verify the change.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfList.js
@@ -530,6 +530,17 @@ define(["dojo/_base/declare",
       _filterDelay: 1000,
 
       /**
+       * The key of the item to attempt to focus on in the loaded results. This can be set via the
+       * payload handled by [onReloadData]{@link module:alfresco/lists/AlfList#onReloadData}.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.77
+       */
+      _focusItemKey: null,
+
+      /**
        * This timeout pointer is used to ensure the loading message doesn't display forever
        *
        * @instance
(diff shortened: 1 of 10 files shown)


aikau/src/main/resources/alfresco/lists/AlfList.js
```diff
@@ -1242,11 +1259,20 @@ define(["dojo/_base/declare",
        * [loadData]{@link module:alfresco/lists/AlfList#loadData}.
        *
        * @instance
+       * @param {object} [payload] The payload supplied when making the reload request.
+       * @param {string} [payload.focusItemKey] An item to focus on if it is in the data that is reloaded
        */
-      onReloadData: function alfresco_lists_AlfList__onReloadData() {
+      onReloadData: function alfresco_lists_AlfList__onReloadData(payload) {
```
MartinDoyleUK: If this is a payload, it suggests a publication. If that is the case then the parameters should be defined against the event. If it's not the case then we should avoid calling it a payload to avoid confusion.

aikau/src/main/resources/alfresco/lists/views/AlfListView.js
```diff
@@ -235,6 +235,40 @@ define(["dojo/_base/declare",
...
+      focusOnItem: function alfresco_lists_views_AlfListView__focusOnItem(itemKey) {
+         if (this.docListRenderer && this.docListRenderer._renderedItemWidgets)
+         {
+            array.some(this.docListRenderer._renderedItemWidgets, function(widgets) {
+               return array.some(widgets, function(widget) {
+                  var found = false;
+                  if (widget && 
+                      widget.currentItem && 
+                      (widget.currentItem[this.itemKey] || widget.currentItem[this.itemKey] === 0) &&
+                      widget.currentItem[this.itemKey].toString() === itemKey)
```
MartinDoyleUK: Would `lang.getObject()` make this any more concise??

draperd: No.

aikau/src/main/resources/alfresco/lists/AlfList.js
draperd: I've updated topics with this information.

PR #1165, opened 2016-07-25
Correct German short date based on DIN 5008 (allowed alt to ISO 8601)

AFaust:
For whatever reason Alfresco has been using an incorrect German locale short date format for ages in Share and this has been transported to Aikau as well. This PR corrects the format based on DIN 5008's revision of 2006 which re-allows the traditional (d)d.(m)m.(yy)yy format as an alternative to the officially recommended DIN ISO 8601 yyyy-mm-dd.
See https://en.wikipedia.org/wiki/Date_and_time_notation_in_Europe and https://en.wikipedia.org/wiki/Date_format_by_country for reference


davidcognite:
Hey Axel - thanks for raising this, I'll flag it with our localisation manager for review, as we'd need to update our translation memories (across products).

gbroadbent:
Hello David,

What is this?

Best regards,

Gloria

davidcognite:
I've discussed this with @gbroadbent and have raised https://issues.alfresco.com/jira/browse/ALF-21703 to be fixed when Aikau and Share are next translated.

I'll close this PR as localised files are maintained outside of this GitHub repo, but thanks for raising it - it's always good to get feedback on the translations from native speakers of those languages.

PR #1166, opened 2016-07-25
Add support for button payloads to be dynamically constructed

AFaust:
Constructing the publication payload of a publication-enabled widget based on the "current item" associated with a widget sub-structure is an almost ubiquitous pattern in Aikau, yet it is not supported by the AlfButton module. This PR simply adds this missing support, retaining existing behaviour for all users of AlfButton due to the default payload type being CONFIGURED within _PublishPayloadMixin.

### DIFF
--- a/aikau/src/main/resources/alfresco/buttons/AlfButton.js
+++ b/aikau/src/main/resources/alfresco/buttons/AlfButton.js
@@ -32,18 +32,20 @@
  * @module alfresco/buttons/AlfButton
  * @extends module:dijit/form/Button
  * @mixes module:alfresco/core/Core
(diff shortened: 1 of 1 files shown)


draperd:
This is a great addition... I'm going to want to add a supporting unit test to verify and prevent future regressions. So I'll check this out and update (which means it can't go in straight-away). This PR is also aimed at master rather than develop, however - you won't need to change this as I'll create a new PR with the unit tests added.

draperd:
I've created https://issues.alfresco.com/jira/browse/AKU-1045 to track the creation of the unit test

draperd:
Replaced by #1183

PR #1167, opened 2016-07-25
Allow non-truthy values to be mapped in AlfDynamicPayloadButton

AFaust:
During the Aikau-ification of the JavaScript Console addon I encountered an issue where the AlfDynamicPayloadButton would not assume the value I was explicitly providing in my payload. The cause of this was the non-truthy nature of the value (explicit null-value as reset of a previously mapped value).
This PR changes the check for data mapping to ensure the data is explcitly provided but in a way that no longer requires the mapped to be truthy. Other non-truthy variants also include very valid data (boolean false, numerical zero value, empty string) and would be supported after this fix.

### DIFF
--- a/aikau/src/main/resources/alfresco/buttons/AlfDynamicPayloadButton.js
+++ b/aikau/src/main/resources/alfresco/buttons/AlfDynamicPayloadButton.js
@@ -196,9 +196,9 @@ define(["dojo/_base/declare",
             if (dataMapping.hasOwnProperty(key))
             {
                // Check that the data actually exists...
-               var value = lang.getObject(key, false, data);
-               if (value)
+               if (lang.exists(key, data))
                {
+                  var value = lang.getObject(key, false, data);
                   // ...and then set it as requested
                   lang.setObject(dataMapping[key], value, this.publishPayload);
                }


draperd:
I think this needs a supporting unit test, which I will add as soon as I can. Although the PR is aimed at master I will need to create a new PR with the tests - so you won't need to change this one.

draperd:
Do you have a simple JSON model example of how you'd like this change to be used? It will make it easier for me to create the unit test - it also means that we'll be testing the right sort of thing.

AFaust:
In my JavaScript Console Aikau-ification branch I use a dynamic payload button to collect parameters from various inputs (i.e. codemirror editors). Simplified this could be represented as (jsconsole/button/DynamicPayloadButton is a sub-module of AlfDynamicPayloadButton with the changes suggested in the PR):

```javascript
model.jsonModel.widgets = [{
            name : 'jsconsole/button/DynamicPayloadButton',
            config : {
                label : 'Execute',
                publishTopic : 'EXECUTE_JS',
                publishPayload : {},
                publishPayloadSubscriptions : [ {
                    topic : 'UPDATE_JS_SOURCE',
                    dataMapping : {
                        content : 'javaScriptSource',
                        selectedContent : 'selectedJavaScriptSource'
                    }
                }]
            }
        }, {
            name : 'alfresco/buttons/AlfButton',
            config : {
                label : 'setSource (with selected code)',
                publishTopic : 'UPDATE_JS_SOURCE',
                publishPayload : {
                        content: 'print(10);',
                        selectedContent: 'print(10);'
                }]
            }
        }, {
            name : 'alfresco/buttons/AlfButton',
            config : {
                label : 'removeCodeSelection',
                publishTopic : 'UPDATE_JS_SOURCE',
                publishPayload : {
                        content: 'print(10);',
                        selectedContent: null
                }]
            }
        }];
```

The ```selectedJavaScriptSource``` is an imported part of the execution trigger payload as it will result in only the selected JS code being executed by the script backend. It will be frequently set and reset depending on selection in the codemirror editor instances. When ```EXECUTE_JS``` is published, the ```selectedJavaScriptSource``` must always reflect the state of the last ```UPDATE_JS_SOURCE``` payload.

draperd:
Thanks for providing the example model... I've created https://issues.alfresco.com/jira/browse/AKU-1052 to address this and will try and get it resolved as soon as I can.

draperd:
This has now been super-ceded by #1195

PR #1168, opened 2016-07-25
Add locale generalization so TinyMCE actually uses client locale

AFaust:
The languages supported by TinyMCE editor are maintained as two-letter country codes while the JS_LOCALE will typically be the combination of country and language code, i.e. de_DE. The current locale matching in the TinyMCE editor does not detect / select an available, generalized locale and essentially always falls back to English.
This PR adds locale generalization checks during in the locale match code of the TinyMCE editor.


draperd:
This is PR is aimed at master rather than develop. Could you change it please. It should be sufficient for me to just run the existing unit test to verify, I don't think we'll need a new unit test here.

AFaust:
Closed due to inability in GitHub to change target branch of a PR

PR #1169, opened 2016-07-31
Use site for loading of actions via Xhr and allow client specifying view

AFaust:
When using the XhrActions module (as it is done in the faceted search page) some actions may not be listed as "available" if
- the  document / item is located within a site, and
- uses an evaluator based on "evaluator.doclib.action.groupMembership", and
- data is requested in a non-raw format
This is due to the fact that the code for handling XhrActions and retrieval of single document data is not handling the site of the document / item.

This PR addresses the issue above by adding site handling to the XhrActionsMixin and the DocumentService modules to ensure the call to the Share-tier data web script is passed the specific site of the document / item. Additionally, it allows a client module to request a specific view instead of the default "browse" (e.g. if actions are associated with a different action group than XY-browse, i.e. XY-details).

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
@@ -46,6 +46,29 @@ define(["dojo/_base/declare",
        * @since 1.0.46
        */
       i18nRequirements: [{i18nFile: "./i18n/_XhrActionsMixin.properties"}],
+      
+      /**
+       * The lookup key into current item to retrieve the short name of the site in which the current item is
+       * located. This may be relevant to correctly obtain all item actions if evaluation of these actions is
+       * dependant on the site scope.
+       * 
+       * @instance
+       * @type {string}
+       * @default null
+       * @since 1.0.8x
+       */
+      currentItemSiteShortNameKey : null,
+      
+      /**
+       * The name of the view for which to load actions. Depending on the backend loading actions this may
+       * allow a differentiation between various scopes that actions have been associated with.
+       * 
+       * @instance
+       * @type {string}
+       * @default null
+       * @since 1.0.8x
+       */
+      actionView : null,
 
       /**
        * Overrides the [inherited function]{@link module:alfresco/renderers/_ActionsMixin#createDropDownMenu}
@@ -99,13 +122,28 @@ define(["dojo/_base/declare",
          var nodeRef = lang.getObject("nodeRef", false, this.currentItem);
          if (nodeRef)
          {
+            // depending on how item was loaded site can be coded in different ways
+            var site;
+            if (this.currentItemSiteShortNameKey)
+            {
+                site = lang.getObject(this.currentItemSiteShortNameKey, false, this.currentItem);
+            }
+            else
+            {
+                // default fallbacks for search and doclib data web scripts (JsNode vs. bare)
+                site = site || lang.getObject("site.shortName", false, this.currentItem);
+                site = site || lang.getObject("node.location.site.name", false, this.currentItem);
+                site = site || lang.getObject("location.site.name", false, this.currentItem);
+            }
             // Generate a UUID for the response to the publication to ensure that only this widget
             // handles to the XHR data...
             var responseTopic = this.generateUuid();
             this._xhrDataRequestHandle = this.alfSubscribe(responseTopic + "_SUCCESS", lang.hitch(this, this.onXhrData, callback), true);
             this.alfPublish(topics.GET_DOCUMENT, {
                alfResponseTopic: responseTopic,
-               nodeRef: nodeRef
+               nodeRef: nodeRef,
+               site : site,
+               view : this.actionView
             }, true);
          }
          else
--- a/aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
@@ -54,8 +54,7 @@ define(["dojo/_base/declare",
        * 
        * @instance
        * @type {string}
-       * @default null
-       * @since 1.0.8x
+       * @since 1.0.79
        */
       currentItemSiteShortNameKey : null,
       
@@ -65,8 +64,7 @@ define(["dojo/_base/declare",
        * 
        * @instance
        * @type {string}
-       * @default null
-       * @since 1.0.8x
+       * @since 1.0.79
        */
       actionView : null,
 
--- a/aikau/src/main/resources/alfresco/services/DocumentService.js
+++ b/aikau/src/main/resources/alfresco/services/DocumentService.js
@@ -187,7 +187,7 @@ define(["dojo/_base/declare",
             // View mode and No-cache
             var params = "?view=";
             params += encodeURIComponent(payload.view || "browse");
-    		params += "&noCache=" + new Date().getTime() + "&includeThumbnails=true";
+            params += "&noCache=" + new Date().getTime() + "&includeThumbnails=true";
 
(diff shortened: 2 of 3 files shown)


aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
```diff
@@ -99,13 +122,28 @@ define(["dojo/_base/declare",
          var nodeRef = lang.getObject("nodeRef", false, this.currentItem);
          if (nodeRef)
          {
+            // depending on how item was loaded site can be coded in different ways
+            var site;
+            if (this.currentItemSiteShortNameKey)
```
AFaust: This configuration option was put in place to allow short-circuit of the "default lookup chain" if a developer is sure enough about how site can be extracted from the current item.

aikau/src/main/resources/alfresco/services/DocumentService.js
```diff
@@ -184,21 +184,10 @@ define(["dojo/_base/declare",
             var nodeRef = NodeUtils.processNodeRef(payload.nodeRef),
             targetNodeUri = nodeRef.uri;
 
-            // Construct the URI for the request...
```
AFaust: This code part was a remnant from a copy&paste from onRetrieveDocumentsRequest, but the calculated uriPart was not used at all.

aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
```diff
@@ -46,6 +46,29 @@ define(["dojo/_base/declare",
...
       i18nRequirements: [{i18nFile: "./i18n/_XhrActionsMixin.properties"}],
+      
+      /**
+       * The lookup key into current item to retrieve the short name of the site in which the current item is
+       * located. This may be relevant to correctly obtain all item actions if evaluation of these actions is
+       * dependant on the site scope.
+       * 
+       * @instance
+       * @type {string}
+       * @default null
```
draperd: When the initial assignment is `null`, a String, number or a boolean it is not necessary to set it on the `default` annotation... JSDoc automatically figures that out.

aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
```diff
@@ -46,6 +46,29 @@ define(["dojo/_base/declare",
...
+       * @since 1.0.8x
```
draperd: If you change this to 1.0.79 I'll look to merge this for the next release

aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
```diff
@@ -46,6 +46,29 @@ define(["dojo/_base/declare",
...
+       */
+      currentItemSiteShortNameKey : null,
+      
+      /**
+       * The name of the view for which to load actions. Depending on the backend loading actions this may
+       * allow a differentiation between various scopes that actions have been associated with.
+       * 
+       * @instance
+       * @type {string}
+       * @default null
```
draperd: As with comment above, can you remove the `null` please?

aikau/src/main/resources/alfresco/renderers/_XhrActionsMixin.js
```diff
@@ -46,6 +46,29 @@ define(["dojo/_base/declare",
...
+       * @since 1.0.8x
```
draperd: And as above, could you set this to 1.0.79 please?

aikau/src/main/resources/alfresco/services/DocumentService.js
draperd: Thanks for highlighting this and removing it - I'd not spotted it was not being used.

aikau/src/main/resources/alfresco/services/DocumentService.js
```diff
@@ -208,7 +197,12 @@ define(["dojo/_base/declare",
             }
             else
             {
-               url = AlfConstants.URL_SERVICECONTEXT + "components/documentlibrary/data/node/" + targetNodeUri + params;
+               url = AlfConstants.URL_SERVICECONTEXT + "components/documentlibrary/data/";
+               if (payload.site)
+               {
+                   url += 'site/' + encodeURIComponent(payload.site) + '/';
```
draperd: Could you change the single quotes to be double quotes please to avoid JSHint errors on out automated builds

draperd:
Thanks for raising this... I've made a few initial comments and I'll checkout your branch and run the tests locally to ensure there are no failures.

AFaust:
Requested changes have been made to PR branch.

draperd:
Thanks for making those changes - I'm running the unit tests and hitting some issues, I'll add an update when I've got to the bottom of what the problems/failures are.

draperd:
I've tested the changes on Share and can't find any issues, I think it might be something to do with the mock data provided for the unit tests, the URLs must have changed somehow - I'll dig further into it.

aikau/src/main/resources/alfresco/services/DocumentService.js
```diff
@@ -329,7 +325,12 @@ define(["dojo/_base/declare",
          }
          else
          {
-            url = AlfConstants.URL_SERVICECONTEXT + "components/documentlibrary/data/doclist/" + params;
+            url = AlfConstants.URL_SERVICECONTEXT + "components/documentlibrary/data/";
+            if (payload.site)
```
draperd: It looks like this `if` block is not required. Stepping through the code in the unit test it seems that we're getting the site data included twice, for example on the "develop" branch we see this request:

```
/aikau/service/components/documentlibrary/data/doclist/all/site/site1/documentlibrary?filter=path&size=25&pos=1&sortAsc=false&sortField=cm%3Aname&view=browse&noCache=1470142533664
```

...but on your branch we see this:

```
/aikau/service/components/documentlibrary/data/site/site1/doclist/all/site/site1/documentlibrary?filter=path&size=25&pos=1&sortAsc=false&sortField=cm%3Aname&view=browse&noCache=1470142609664
```

Note the additional "site/site1" after "data" and before "doclist". Any thoughts on this?

I'll continue to test my update against Share to see if it causes any issues.

draperd:
I think if you remove that `if` block mentioned then this will be good to merge.

aikau/src/main/resources/alfresco/services/DocumentService.js
AFaust: The fragment until (and including) data/site/site1 is consumed by the Share data web script surf-doclist.get (URL pattern /components/documentlibrary/data/site/{site}/{webscript}/{params}) while the remainder is captured as the repository web script to call. As a result, the site is contained twice if both tiers are to function correctly.

draperd: OK... I hadn't realised that. Well, it was working in Share, so I guess that makes sense... in which case, it will be necessary to push some changes to the mocking service (aikau/src/test/resources/testApp/js/aikau/testing/mockservices/FullDocLibMockXhr.j).

In the `setupServer` function the first call to `this.server.respondWith` needs to be updated.

So...

```JavaScript
/\/aikau\/service\/components\/documentlibrary\/data\/doclist\/all\/site\/site1\/documentlibrary\?filter=path(.*)/,
```

...needs to become...

```JavaScript
/\/aikau\/service\/components\/documentlibrary\/data\/site\/site1\/doclist\/all\/site\/site1\/documentlibrary\?filter=path(.*)/,
```

draperd: The last call (for favourites) will also need updating!

draperd:
MERGED ! 

Thanks for your patience with this one and thanks for the contribution!

PR #1170, opened 2016-07-31
Correct I18n/encoding handling of page title, esp. prefix

AFaust:
This PR addresses some issues when trying to use the alfresco/header/Title module for customizing the title of the page. For one, the prefix and label for the browser title were previously not looked up before actually being used during page initialisation. This resulted in a message key being included in the browser title bar unless the "updatePageTitle" pubSub was triggered. Additionally, the HTML encoding of the prefix during postMixInProperties was redundant and caused character sequences like "&Auml;" to appear in the browser title bar when a German umlaut or other non-ascii/ansi characters are contained in the localized message.

Edit: Forgot to select "develop" as the target branch - hope this can be changed before merging...

### DIFF
--- a/aikau/src/main/resources/alfresco/header/Title.js
+++ b/aikau/src/main/resources/alfresco/header/Title.js
@@ -132,22 +132,23 @@ define(["dojo/_base/declare",
        * @instance
        */
       postMixInProperties: function alfresco_header_Title__postMixInProperties() {
+         if (this.browserTitlePrefix)
+         {
+            this.browserTitlePrefix = this.message(this.browserTitlePrefix);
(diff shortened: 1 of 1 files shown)


draperd:
This PR is aimed at "master" rather than "develop".

AFaust:
I actually thought I remembered to pick develop on most of these PRs, but turns out only a few ended up correct. Unfortunately GitHub hasn't got its act together for more than 3 years now (https://github.com/isaacs/github/issues/18) and I have to re-create all affected PRs. Looks like any branching model that uses a non-master branch for contributions is a real pita on this platform...

PR #1174, opened 2016-07-31
Respect explicitly empty action arrays (i.e. lack of permissions)

AFaust:
Any array of item-level actions, either explicitly define or loaded from  a backend, must be respected by the alfresco/renderer/Actions module (or any of its derivatives). An empty array of item-level actions loaded from the Share data web script may be the result of a lack of permissions by the current user or simply an action set configured as empty by a developer / administrator in the Share configuration. 

Currently, the ActionsMixin will fallback on the default "widgetsForActions" when
- mergeActions is set to false, and
- customActions is either not defined or an empty array, and
- currentItem.actions is either not defined or an empty array

The  point of either configuring an explicit empty array for customActions or configuring Share actions in a way that may result in an emty item-level action list is to ensure the user can't perform any action on the item if pre-conditions are not met. Ignoring this valid and explicit state and falling back to a (default) widgetsForActions config is not appropriate.

This PR changes the behaviour to always respect the presence of an actions array even if it may have a length of zero.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
@@ -41,9 +41,10 @@ define(["dojo/_base/declare",
         "dojo/_base/lang",
         "service/constants/Default",
         "alfresco/core/ArrayUtils",
-        "alfresco/core/JsNode"],
+        "alfresco/core/JsNode",
+        "alfresco/core/ObjectTypeUtils"],
         function(declare, AlfCore, CoreWidgetProcessing, _AlfDocumentListTopicMixin, _PublishPayloadMixin, ObjectProcessingMixin, 
-                 _AlfPopupCloseMixin, AlfMenuItem, array, lang, AlfConstants, AlfArray, JsNode) {
+                 _AlfPopupCloseMixin, AlfMenuItem, array, lang, AlfConstants, AlfArray, JsNode, ObjectTypeUtils) {
 
    return declare([AlfCore, CoreWidgetProcessing, _AlfDocumentListTopicMixin, _PublishPayloadMixin, ObjectProcessingMixin, _AlfPopupCloseMixin], {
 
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
```diff
@@ -192,19 +192,19 @@ define(["dojo/_base/declare",
          if (this.mergeActions === true)
          {
             // Add the actions on the currentItem and then add additional custom actions...
-            this.currentItem.actions && array.forEach(this.currentItem.actions, lang.hitch(this, this.addAction));
-            this.customActions && array.forEach(this.customActions, lang.hitch(this, this.addAction));
+            lang.isArray(this.currentItem.actions) && array.forEach(this.currentItem.actions, lang.hitch(this, this.addAction));
```
AFaust: lang.isArray used to avoid errors due to non-array truthy values that may be returned / configured

aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
```diff
@@ -220,7 +220,6 @@ define(["dojo/_base/declare",
        * @instance
        */
       processWidgetsForActions: function alfresco_renderers__ActionsMixin__processWidgetsForActions() {
-         // TODO: We probably want to avoid rendering all the actions until the menu is opened...
```
AFaust: This comment is no longer applicable as far as I understand due to AKU-909

draperd:
Thanks for raising all these pull requests, I wasn't ignoring them... I've only just come back from vacation. However, I've noticed that a few of them (like this one) are pull requests to "master" rather than "develop" (I've inadvertently merged one, which I'll need to correct)... can you make sure that they're all for "develop" please?

For some of them, I'll probably need to add supporting tests so they might take a little while to be merged in.

aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
draperd: As a general rule, we use the "isArray" function of `alfresco/core/ObjectTypeUtils` (which is an alias for lang.isArray). The reason being is that it gives us the option at a future point to swap out the implementation for a more performant solution or even to go "native" should such as feature ever be available across all supported browsers.

However, I'm not going to hold this PR for that change.

draperd:
This PR looks good (one minor comment on the use of lang.isArray rather than the alias in ObjectTypeUtils)... however, it's still targeted at master rather than develop. If you can recreate this PR against develop then I'd be happy to merge as I've verified it locally against the unit tests.

aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
AFaust: Ok - I was trying to avoid adding another dependency to the module, but can certainly add ObjectTypeUtils.
As far as I am aware, all the type-checking that both lang and ObjectTypeUtils do can already be done in a native way supported across all (supported) browsers (considering IE 8 to be unsupported as it is not supported my MS anymore).

AFaust:
Recreated as #1182

PR #1175, opened 2016-08-01
Correct I18n/encoding handling of page title, esp. prefix

AFaust:
Recreation of #1170 

This PR addresses some issues when trying to use the alfresco/header/Title module for customizing the title of the page. For one, the prefix and label for the browser title were previously not looked up before actually being used during page initialisation. This resulted in a message key being included in the browser title bar unless the "updatePageTitle" pubSub was triggered. Additionally, the HTML encoding of the prefix during postMixInProperties was redundant and caused character sequences like "Ä" to appear in the browser title bar when a German umlaut or other non-ascii/ansi characters are contained in the localized message.

### DIFF
--- a/aikau/src/main/resources/alfresco/header/Title.js
+++ b/aikau/src/main/resources/alfresco/header/Title.js
@@ -132,22 +132,23 @@ define(["dojo/_base/declare",
        * @instance
        */
       postMixInProperties: function alfresco_header_Title__postMixInProperties() {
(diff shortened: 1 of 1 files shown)


draperd:
Thanks for recreating this against develop. I've checked it out locally and verified that the unit tests continue to pass. I don't think that we need additional tests here, so I'll merge. Thanks again for raising this.

PR #1176, opened 2016-08-01
Locale generalization so JS_LOCALE of de_DE results in use of de

AFaust:
Recreation of #1168 

The languages supported by TinyMCE editor are maintained as two-letter country codes while the JS_LOCALE will typically be the combination of country and language code, i.e. de_DE. The current locale matching in the TinyMCE editor does not detect / select an available, generalized locale and essentially always falls back to English.
This PR adds locale generalization checks during in the locale match code of the TinyMCE editor.

### DIFF
--- a/aikau/src/main/resources/alfresco/editors/TinyMCE.js
+++ b/aikau/src/main/resources/alfresco/editors/TinyMCE.js
@@ -226,14 +226,23 @@ define(["dojo/_base/declare",
          // Check that the language requested is supported...
          if (config.language) {
             var locales = this.supportedLocales.split(",");
-            var locale = "en";
+            var locale, bestGeneralizedLocale;
             for (var i = 0, j = locales.length; i < j; i++) {
                if (locales[i] === config.language) {
                   locale = config.language;
                   break;
                }
+               
+               if (config.language.indexOf(locales[i]) === 0)
+               {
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/editors/TinyMCE.js
```diff
@@ -226,14 +226,23 @@ define(["dojo/_base/declare",
...
+               {
+                   if (bestGeneralizedLocale === undefined || locales[i].length > bestGeneralizedLocale.length)
+                   {
+                       bestGeneralizedLocale = locales[i];
+                   }
+               }
             }
-            config.language = locale;
+            
+            config.language = locale || bestGeneralizedLocale || 'en';
```
draperd: We have JSHint on our build servers to use double rather than single quotes (the build aggressively fails on JSHint errors!). I'm flagging this up for future reference more than anything else as I'm happy to merge myself and then fix the build failure. But if it's simple enough for you to push a quick update then I'd be very grateful.

draperd:
Thanks for recreating this PR...I've made one minor comment on our preferred JSHint configuration, but otherwise I'm happy to merge this having tested it out locally.

draperd:
Thanks for making that update so quickly - I've now merged the request. It will be in the 1.0.79 release which will either be later this week or early next week.

PR #1182, opened 2016-08-01
Respect explicitly empty action arrays (i.e. lack of permissions)

AFaust:
Recreation of #1174 

Any array of item-level actions, either explicitly define or loaded from a backend, must be respected by the alfresco/renderer/Actions module (or any of its derivatives). An empty array of item-level actions loaded from the Share data web script may be the result of a lack of permissions by the current user or simply an action set configured as empty by a developer / administrator in the Share configuration.

Currently, the ActionsMixin will fallback on the default "widgetsForActions" when

-mergeActions is set to false, and
-customActions is either not defined or an empty array, and
-currentItem.actions is either not defined or an empty array

The point of either configuring an explicit empty array for customActions or configuring Share actions in a way that may result in an emty item-level action list is to ensure the user can't perform any action on the item if pre-conditions are not met. Ignoring this valid and explicit state and falling back to a (default) widgetsForActions config is not appropriate.

This PR changes the behaviour to always respect the presence of an actions array even if it may have a length of zero.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
+++ b/aikau/src/main/resources/alfresco/renderers/_ActionsMixin.js
@@ -41,9 +41,10 @@ define(["dojo/_base/declare",
         "dojo/_base/lang",
         "service/constants/Default",
         "alfresco/core/ArrayUtils",
-        "alfresco/core/JsNode"],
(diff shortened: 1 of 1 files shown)


draperd:
Merged - thanks for making the requested changes!

PR #1189, opened 2016-08-03
AKU-1047: Support disabled state in PushButtons

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1047 to ensure that PushButtons can't be used when they're in the disabled state. Unit tests have been updated to verify this change. This doesn't change the disabled appearance as this has been raised by a separate issue.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
@@ -418,6 +418,18 @@ define(["dojo/_base/declare",
          {
             this.wrappedWidget.set("disabled", status);
          }
+         if (this.domNode)
+         {
+            if (status)
+            {
+               domClass.add(this.domNode, "alfresco-forms-controls-BaseFormControl--disabled");
+            }
+            else
+            {
+               domClass.remove(this.domNode, "alfresco-forms-controls-BaseFormControl--disabled");
+            }
+         }
+         
          this.validate();
       },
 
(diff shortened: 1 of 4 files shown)


aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
```diff
@@ -54,6 +56,28 @@ define(["alfresco/core/CoreWidgetProcessing",
...
+       * @instance
+       * @param  {boolean} status Whether or not the control should be disabled
+       * @since 1.0.80
+       */
+      alfDisabled: function alfresco_forms_controls_BaseFormControl__alfDisabled(status) {
+         this.inherited(arguments);
+
+         if (this.wrappedWidget && this.wrappedWidget.opts)
+         {
+            array.forEach(this.wrappedWidget.opts, function(option) {
```
davidcognite: Do we need to use the lib method here? Can't we use array.prototype.forEach instead?

aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
```diff
@@ -418,6 +418,18 @@ define(["dojo/_base/declare",
...
+            if (status)
+            {
+               domClass.add(this.domNode, "alfresco-forms-controls-BaseFormControl--disabled");
+            }
+            else
+            {
+               domClass.remove(this.domNode, "alfresco-forms-controls-BaseFormControl--disabled");
+            }
+         }
+
```
davidcognite: I have this exact code in my local branch for a PR I started working on last night!

aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
draperd: I've had this conversation with Martin before... we don't have to, but for the time being I want to keep the code consistent throughout the application. I'm also just in the habit of writing the code like this, so it's second nature.

davidcognite: Not a problem at all. I just get excited that we can now use ES5!

Issue #1185, opened 2016-08-01
100% node.js CPU usage during grunt local_tests without any progress


AFaust:
After another attempt to set up a Linux-based VM for testing Aikau following https://github.com/Alfresco/Aikau/wiki/Aikau%20Setup%20%28Linux%29 I am again stumped about how unstable the node.js based setup is. (Although, nothing with regards to "quality" or "professionalism" of node.js or its ecosystem is capable of disappointing me much further at this point.)

I basically executed every step to the letter up to "Running the tests". Instead of Mint 17.1 I used the newer 17.3. Since google-chrome-stable could not be found via "apt-cache search" I opted for chromium-browser instead. The node.js version installed by nvm is 6.3.1.

When the first "grunt local_tests" is executed as listed under "Running the tests" (no changes made to Suites.js), the node.js process initially runs fine until the intern task is being processed. After clearing the console screen it remains empty. "top" reports continuous 100% CPU usage by the node.js process from then on until I Ctrl-c the grunt process and then kill the still-running node.js process.
The only indication that a test might actually be attempted to be run is in a single log line output from the Jetty test server run via "mvn clean install jetty:run" in another shell. The line reads:
ERROR [extensions.surf.DependencyAggregator] Could not find compressed file: js/aikau/testing/RequireEverything.js

When I previously attempted to set up a test VM on the same Mint version about 4 months ago, I at least got some tests to run intermittently (most tests though would fail to be properly opened in the browser, some reported seemingly trivial/stupid issues though tests were run against unaltered master or develop), albeit at some point the same 100% CPU usage situation settled in. Before that I attempted to set up on my local Windows 7 environment, but the intern NPM module failed to be installed with cryptic/unspecific internal errors from associated installation scripts (support/fixdeps.js).

draperd:
OK, thanks for such a detailed explanation of the problems. I'll see if I can get to the bottom of the situation. I know that the unit tests can be quite intensive - what resources are you giving to your VM (number of CPUS, memory, etc)?

AFaust:
4 CPUs, 8 GiB memory (you snooze - you loose). 100% CPU reported by "top" means 1 of 4 CPU is occupied. No other processes (i.e. Java / Jetty) seemed to take any resources while I observed the state via "top"

draperd:
OK... I've created a similar VM with Mint and will attempt to step through the setup instructions we've provided and see what happens. I probably won't get around to this until tomorrow though.

[...]

PR #1191, opened 2016-08-03
Paginator pageBack / pageForward is not setting page label via publication payload

AFaust:
When the pageBack and pageForward menu items of the Paginator are used, the payload is lacking the ```label``` attribute that the menu items in the page drop down include. This results in the ```value``` of the payload being used as the page label (briefly) until processLoadedDocuments corrects this again. This leads to an unstable layout as text grows / shrinks which users may perceive negatively. Additionally, it is technically possible for the user to trigger multiple pageBack / pageForward events before the first load request has been completed (and the pageBack/pageForward items can be disabled), which may cause the value of ```currentPage``` to exceed the valid value range / page count bounds.

This PR adds the precaclulated labels to pageBack / pageForward and a guard to avoid ```currentPage``` exceeding the value range. Due to #1185 I have not run the unit tests but have tested this branch manually via the Aikau Unit Testing page.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/Paginator.js
+++ b/aikau/src/main/resources/alfresco/lists/Paginator.js
@@ -377,10 +377,23 @@ define(["dojo/_base/declare",
        */
       onPageBack: function alfresco_lists_Paginator__onPageBack(payload) {
          // jshint unused:false
-         this.currentPage--;
-         this.alfPublish(this.pageSelectionTopic, {
-            value: this.currentPage
-         });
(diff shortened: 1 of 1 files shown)


draperd:
I've done a quick run through a selection of obvious tests that might fail and haven't hit any issues. I've also tried it locally on Share and it seems fine. The code looks good. I'll run a full regression suite and if it passes (which I'm sure it will) then I'll merge.

draperd:
Full regression test passed, merging.

PR #1192, opened 2016-08-03
AKU-1049: Handle late form field registration

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1049 to ensure that form fields that are late to register (in this case when added via TabbedControls that do not render immediately) that the form re-publishes all the other field values to give it the late registering field the opportunity to correctly evaluate any dynamic rules (for visibility, disablement and requirement). A unit test based on the real-world RM use case has been provided.


davidcognite:
LGTM

PR #1193, opened 2016-08-03
AKU-1048: Set a themeable opacity on disabled form fields

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1048 to apply a themeable opacity to disabled form fields. No unit tests have been added for this largely cosmetic change.


davidcognite:
LGTM. Nice use of LESS var & builds on disabled state added in #1189

PR #1194, opened 2016-08-03
AKU-1051: Optional PushButtons label when no options

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1051 to add support for an optional label that can be displayed when no options are available for the PushButtons control. A unit test has been added to verify that the label is displayed.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
@@ -38,8 +38,9 @@ define(["alfresco/core/CoreWidgetProcessing",
         "dojo/_base/lang",
         "dojo/dom-attr",
         "dojo/dom-class",
+        "dojo/dom-construct",
         "alfresco/forms/controls/PushButtonsControl"],
-       function(CoreWidgetProcessing, BaseFormControl, declare, array, lang, domAttr, domClass) {
+       function(CoreWidgetProcessing, BaseFormControl, declare, array, lang, domAttr, domClass, domConstruct) {
 
    return declare([BaseFormControl, CoreWidgetProcessing], {
 
@@ -54,6 +55,16 @@ define(["alfresco/core/CoreWidgetProcessing",
        * @since 1.0.65
        */
       firstValueIsDefault: false,
+
+      /**
+       * An optional message to display when there are no buttons available for display.
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.80
+       */
+      noButtonsLabel: null,
 
       /**
        * Extends the [inherited function]{@link module:alfresco/forms/controls/BaseFormControl#alfDisabled}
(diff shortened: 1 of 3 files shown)


aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
```diff
@@ -144,6 +155,30 @@ define(["alfresco/core/CoreWidgetProcessing",
...
+      setOptions: function alfresco_forms_controls_BaseFormControl__setOptions(options) {
+         this.inherited(arguments);
+
+         if (options && 
+             options.length === 0 && 
+             this.noButtonsLabel && 
+             this.wrappedWidget)
+         {
+            var span = domConstruct.create("span", {
+               "class": "alfresco-forms-controls-PushButtons__noOptionsLabel"
```
MartinDoyleUK: You've probably already verified this works, but shouldn't this be `className`? Maybe Dojo handles it more intelligently than pure JS!

aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
```diff
@@ -144,6 +155,30 @@ define(["alfresco/core/CoreWidgetProcessing",
...
+            }, this.wrappedWidget.domNode);
+            span.appendChild(document.createTextNode(this.message(this.noButtonsLabel)));
```
MartinDoyleUK: Quicker (more efficient??) than creating the text node manually, one could just write:

```javascript
span.textContent = this.message(this.noButtonsLabel);
```

aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
draperd: Yes, this works... note that it's being looked for in the unit test

aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
draperd: Was thinking about XSS prevention... I thought that this was the approach we'd used previously

MartinDoyleUK: Using textContent is as safe as createTextNode and is shorter. There's no particular reason to use one over the other, unless you specifically need a handle on the textNode I guess.

aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
MartinDoyleUK: Sorry. Just surprised! :)

aikau/src/main/resources/alfresco/forms/controls/PushButtons.js
draperd: Good to know, I'll bear that in mind for next time.

PR #1199, opened 2016-08-07
Lazy initialisation and re-use of page selector menu items (Paginator)

AFaust:
When (re-)loading a paginated list, there is significant overhead caused by the Paginator widget module which always destroys and re-creates all the page selection menu items. This is a side-issue found in my analysis preluding #958 that can be addressed without any refactoring around the core widget processing.

This PR adds lazy initialisation and re-use of existing menu items to the page selector menu. When the user clicks the page selector drop down in order to select another page, the hook will check which page menu items have already been created and instantiates the missing remainders. Existing page menu items are only invalidated / destroyed when either the ```documentsPerPage``` setting has changed or the page an item refers to no longer exists or its upper bound has changed (i.e. partially filled last page).
A special case handling for loads potentially initiated by external triggers (i.e. hash change) has been added to the ```processLoadedDocuments``` hook to ensure the current page label is always displayed correctly. This was - as far as I could determine - the primary reason / benefit of the previous behaviour of always re-creating the entire page selector menu.

This PR includes a minor change to AlfCheckableMenuItem to ensure its ```onGroupSelection``` subscription handler is properly handling corner-cases when an item in the same group may be added after the others and should be considered the new, "checked" item via its config.

I tried looking into updating unit tests even though my local test setup is not running. Interestingly there does not seem to be a standalone test of Paginator, only one in the context of documentlibrary. As the tests in that context are quite numerous and I'd be changing them "blindly" (without being able to properly run test), I chose not to.
I did run various manual tests on the Aikau Unit Tests page to ensure the end-user behaviour is correct, which resulted (among fixes to the core PR) in the small change to AlfCheckableMenuItem.

The performance impact / benefit is negligible with lists of only a few pages, but can be dramatic in lists with dozens to thousands of pages.
In one project we passed the ```numberFound``` from our SOLR search as ```totalRecords``` to Aikau UI valueing around a couple 100k results, so Paginator would always destroy and re-create menu items for all pages up to the total number of records (page size was 25). This alone frooze the browser for ~30 seconds. (Limiting the number of page menu items is not included in this particular PR as it could conflict with defined UX.)

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/Paginator.js
+++ b/aikau/src/main/resources/alfresco/lists/Paginator.js
@@ -509,6 +509,8 @@ define(["dojo/_base/declare",
                   
                   domClass.remove(this.pageMarker.domNode, "dijitDisabled dijitMenuItemDisabled");
                }
+               
+               this.checkAndUpdatePageSelectionMenu(this.currentPage);
             }
          }
       },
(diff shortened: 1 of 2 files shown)


draperd:
I'll review and test this later this week - hopefully tomorrow. 

WIth regards to your comments on testing Paginator - it is currently included in 10 different unit test pages and is predominantly tested in the PaginationTest ... so although the test name does not quite match, you can be assured that the widget is extensively included in automated tests. Some widgets make sense to test just by themselves, others need to be tested with other widgets.

draperd:
I've checked out the branch and just running against the "alfresco/documentlibrary/PaginationTest" suite I'm getting an error:

> ==================
> ===== FAILED =====
> ==================
> 
> Chrome v48.0.2564.82 on LINUX
> 
> alfresco/documentlibrary/PaginationTest: "Pagination Tests"
> "Test 50 items per page selection update page selector drop-down label"
> Page selector label not updated correctly after switching to 50 items per page: 1-25 of 243
> 
> Firefox v43.0.4 on LINUX
> 
> alfresco/documentlibrary/PaginationTest: "Pagination Tests"
> "Test 50 items per page selection update page selector drop-down label"
> Page selector label not updated correctly after switching to 50 items per page: 1-25 of 243

I'll take a look into why this test is failing, but it would seem to be related to the proposed change given the nature of the test

draperd:
Yes, manual testing verifies that the automated test is indeed correct... when you change the page size the items per page menu is not being updated - so it should be saying "1-50 of 243" (as it is now showing 50 results) but it continues to say "1-25 of 243".

Therefore this cannot be merged as is. I'm going to close this PR now to prevent accidental merging... if you can correct the code to ensure that the label gets updated correctly then please open a new PR. I'm not averse to the change in principle - but we need to make sure that we don't break existing functionality.

AFaust:
Ah - it appears to be a special case for the first page. I can reproduce this problem only on the first page while on all other pages it works correctly. I will adress this and re-open when it is working.

AFaust:
Ok - the problem is now fixed on the branch, but I can't actually re-open this PR myself (I can only re-create it).

draperd:
Re-opening as requested. I'll check out the branch and test locally again.

aikau/src/main/resources/alfresco/lists/Paginator.js
```diff
@@ -755,6 +767,99 @@ define(["dojo/_base/declare",
...
          delete this.__deferredLoadedDocumentData;
          this._controlsLoaded.resolve();
+      },
+      
+      /**
+       *  Handles initialization and necessary updates of the page selection menu items when the popup is opened.
+       * 
+       *  @instance
+       *  @param {object} position the position of the selection menu popup when opened
+       */
```
draperd: Can you add a @since 1.0.82 annotation here please?

aikau/src/main/resources/alfresco/lists/Paginator.js
```diff
@@ -755,6 +767,99 @@ define(["dojo/_base/declare",
...
+                pageEnd = pageStart + parseInt(this.documentsPerPage, 10) - 1; // Deduct 1 because it's 1 - 25 (not 1 - 26!)
+             }
+             else
+             {
+                // ...for the last page just count up to the last document
+                pageEnd = this.totalRecords;
+             }
+
+             label = this.message("list.paginator.page.label", {0: pageStart, 1: pageEnd, 2: this.totalRecords});
+             menuItem = new AlfCheckableMenuItem({
```
draperd: FYI: I don't expect this to be changed for this PR as it replicates the existing code in the module.... but we're trying to phase out the use of "new-ing" widgets directly like this in favour of calling createWidget (inherited from CoreWidgetProcessing)... an example of this can be found in the AlfBreadcrumbTrail widget. The reason is that it means that it's not necessary to pass across the pubSubScope, etc - which means that child widgets will automatically inherit other attributes (such as parentPubSubScope, dataScope, currentItem, currentMetadata, etc). 

So although these aren't relevant here - just something to bear in mind for future pull requests.

draperd:
The PaginationTests pass, now running a full regression suite just to be sure before merging

aikau/src/main/resources/alfresco/lists/Paginator.js
```diff
@@ -755,6 +767,99 @@ define(["dojo/_base/declare",
...
+          lastPreloadedPage = -1;
+          
+          // Delete the (obsolete) previous page selector group contents...
+          array.forEach(this.pageSelectorGroup.getChildren(), function(widget, index, arr) {
+              var virtPageEnd;
+              
+              // need to destroy every existing menu item if documentsPerPage has changed
+              // or state hasn't been fully initialized yet
+              if (this._pageSelectionDocsPerPage === undefined
+                      || this._pageSelectionDocsPerPage !== this.documentsPerPage
```
draperd: JSHint is complaining about the line breaks here... the preference is to break **after** the `||`

aikau/src/main/resources/alfresco/lists/Paginator.js
AFaust: That's good to know. I've always disliked the new-ing as it often required quite nasty overrides/extensions to adapt when necessary.
I hope the Surf dependency analysis regexes will be enhanced (or overriden by Aikau module) so we don't have to keep adding the widget module IDs used in the code (call to createWidget) to the define-call (also always hated to have been forced to add modules there just for the dependency analysis).

draperd:
Merged - thanks for making the updates as requested and submitting this code improvement!

PR #1202, opened 2016-08-11
AKU-1063: Update SelectedItemStateMixin to publish empty array instead of null selected items

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1063 to ensure that the SelectedItemStateMixin always publishes an empty array rather than just `null`. This ensures that the 
AlfSelectedItemsMenuBarPopup will have it's state properly initialised. The unit tests have been updated to reproduce the original failure and ensure it has been fixed.

This PR has also been updated to address https://issues.alfresco.com/jira/browse/AKU-1064 as it builds upon the same test page changes.

### DIFF
--- a/aikau/src/main/resources/alfresco/menus/AlfMenuBarSelectItems.js
+++ b/aikau/src/main/resources/alfresco/menus/AlfMenuBarSelectItems.js
@@ -118,34 +118,41 @@ define(["dojo/_base/declare",
       handleIconClick: function alfresco_menus_AlfMenuBarSelectItems__handleIconClick(evt) {
          this.alfLog("log", "Icon node clicked");
          
-         // Close the popup if it's open...
-         this.closePopupMenu();
-         if (this._itemsSelected === this._NONE)
+         if (this.get("disabled"))
          {
-            // Select all...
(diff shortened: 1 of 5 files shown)


aikau/src/main/resources/alfresco/menus/AlfMenuBarSelectItems.js
```diff
@@ -118,34 +118,41 @@ define(["dojo/_base/declare",
...
+            // No action required - the widget is disabled...
          }
          else
          {
-            // Select none...
-            this.renderNoneSelected();
-            this.alfPublish(this.notificationTopic, { value: "selectNone" });
+            // Close the popup if it's open...
+            this.closePopupMenu();
+            if (this._itemsSelected === this._NONE)
```
MartinDoyleUK: These four conditions really feel like they should be two blocks in a switch statement? DRY principles, 'n all that.

draperd: Whilst I agree in principle, this is just moving existing code into an else block, rather than new code being added so I felt i was outside the scope of the fix.

PR #1204, opened 2016-08-11
AKU-1068: Update copy/move failure messages

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1068 to update the messages displayed when copy/move actions fail. The actions can actually be reported by the API as being a success (e.g. HTTP 200) despite nothing actually being copied or moved. Also, the messages now take into account multiplicity for selected items action handling, and partial failure message no longer attempt to report the individual failures due to a lack of available data on the XHR response.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
+++ b/aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
@@ -241,14 +241,31 @@ define(["dojo/_base/declare",
        * @param {object} payload
        * @fires module:alfresco/core/topics#DISPLAY_NOTIFICATION
        */
-      onActionSuccess: function alfresco_services_ActionService__onActionSuccess(payload) {
+      onActionSuccess: function alfresco_services_actions_CopyMoveService__onActionSuccess(payload) {
          // jshint unused:false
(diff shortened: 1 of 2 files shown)


aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -241,14 +241,31 @@ define(["dojo/_base/declare",
...
-      onActionSuccess: function alfresco_services_ActionService__onActionSuccess(payload) {
+      onActionSuccess: function alfresco_services_actions_CopyMoveService__onActionSuccess(payload) {
          // jshint unused:false
          var subscriptionHandles = lang.getObject("requestConfig.subscriptionHandles", false, payload);
          if (subscriptionHandles)
          {
             this.alfUnsubscribeSaveHandles(subscriptionHandles);
          }
-         if (payload.response.overallSuccess === true)
+         if (payload.response.successCount === 0)
```
MartinDoyleUK: Just to double-check; the situation cannot occur where `successCount > 0` but `successCount < payload.response.totalResults`?

draperd: Correct.... success count will either be equal to or less than the totalResults... totalResults reflects the number of items asked to be copied/moved

PR #1209, opened 2016-08-19
AKU-1069: Folder links request updated node metadata

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1069 to provide the ability for the InlineEditPropertyLink widget to make a request for the latest metadata for the current item. This is required because changing the name of a folder results in the location metadata changing which cannot be refreshed manually safely. The most reliable approach is to support the ability to reload the data after saving so the most recent changes are displayed on the client. This ensures that linking is correct. The unit tests have been updated to support this change.

### DIFF
--- a/aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
+++ b/aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
@@ -44,6 +44,7 @@ define(["dojo/_base/declare",
         "dojo/text!./templates/InlineEditProperty.html",
         "dojo/_base/lang",
         "dojo/_base/array",
+        "dojo/Deferred",
         "dojo/on",
         "dojo/dom-class",
         "dojo/html",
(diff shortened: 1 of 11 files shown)


aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
```diff
@@ -515,15 +582,49 @@ define(["dojo/_base/declare",
...
+
+      /**
+       * This function is called from [onSaveSuccess]{@link module:alfresco/renderers/InlineEditProperty#onSaveSuccess}
+       * when [refreshCurrentItem]{@link module:alfresco/renderers/InlineEditProperty#refreshCurrentItem} is true
+       * and allows the [currentItem]{@link module:alfresco/core/CoreWidgetProcessing#currentItem} to be updated
+       * with the latest data following the update.
+       * 
+       * @instance
+       * @param {object} payload The success payload
+       * @since 1.0.83
```
MartinDoyleUK: Missing @returns for the promise?

draperd: Yep, good call.

aikau/src/main/resources/alfresco/renderers/InlineEditProperty.js
```diff
@@ -476,6 +541,8 @@ define(["dojo/_base/declare",
        */
       onSave: function alfresco_renderers_InlineEditProperty__onSave(formPayload) {
          /*jshint unused:false*/
+         domClass.add(this.domNode, "alfresco-renderers-InlineEditProperty--updating");
```
MartinDoyleUK: Just a casual aside, this one. Is there any point in using a setTimeout somewhere to be sure this displays?

draperd: Discussed offline. Agreed to leave this for now as it's not been problematic in testing. Also, in the majority of cases we don't actually want it to display as it'll be so fast.

PR #1212, opened 2016-08-26
AKU-1072: Support options for hiding validation in form controls

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1072 to provide configuration options for hiding the validation indicators and messages for form controls. The use case here is when no label attribute is required (the exact use case being where a SimplePicker is the sole form control in a form dialog) and we want to avoid the requirement asterisk, invalid field indicator being displayed, etc. This is an opt-in feature meaning that all existing implementations should not be impacted. Unit tests have been provided to verify this change.


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -479,7 +481,8 @@ define(["dojo/_base/declare",
             contentWidth: payload.contentWidth ? payload.contentWidth : null,
             contentHeight: payload.contentHeight ? payload.contentHeight : null,
             handleOverflow: handleOverflow,
-            fixedWidth: fixedWidth
+            fixedWidth: fixedWidth,
+            noMinWidth: payload.noMinWidth || false
```
MartinDoyleUK: One thing you could consider here that would enforce a boolean for truthy as well as falsy values would be to have `!!payload.noMinWidth`. Also applies to line 729 below.

PR #1213, opened 2016-08-28
Fix redundant vertical scrollbars in FilePicker (and likely other dialogs with handleOverflow=false)

AFaust:
The FilePicker shows redundant vertical scrollbars when the content of the currently active tab exceeds the dynamically calculated height of the dialog content panel. This is the result of incorrect calculation of heights (the content panel is larger than it should actually be based on dialog height and buffer for dialog title + toolbar) and a missing parameter when creating the FilePicker dialog to tell the AlfDialog module that the content (FixedHeaderFooter) already handles any overflow.

This PR addresses this layout issue. It:
- adds the handleOverflow parameter to the dialog payload for FilePicker (value: false)
- adds a missing resize watch on the AlfTabContainer (which might be resized either due external constraints, changes in tab contents or addition / removal / activation of tabs)
- adds missing constraints to the dialog content panel height, restricting it to maximum allowed height - either dynamically calculated or (if contents handle overflow) the externally defined height.

Screenshot of original problem with redundant scroll bars:
aikau-FilePicker-scrolling

Note: No unit tests have been adapted. A new machine will hopefully be delivered in the next 14 days which will be used to retry getting unit tests to work for me.

### DIFF
--- a/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
+++ b/aikau/src/main/resources/alfresco/dialogs/AlfDialog.js
@@ -250,7 +250,7 @@ define(["dojo/_base/declare",
          var simplePanelHeight = null;
          if (this.contentHeight)
          {
-            simplePanelHeight = (parseInt(this.contentHeight, 10) - paddingAdjustment) + "px";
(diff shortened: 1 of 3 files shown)


draperd:
Thanks for raising this. I'll need to run the regression tests before merging, but it looks OK on first review.

draperd:
Regression test passed, merging.

PR #1215, opened 2016-08-28
Picker - fix for null reference error

AFaust:
Working on my JavaScript Console Aikau-ification project I encountered a reference error in the getValue function of the Picker module. This PR adds a simple guard for the initialisation of the pickedItemsWidget, which is already checked in a similar way in setValue


draperd:
Thanks for spotting and fixing this - you're quite right that the code should have been more defensive in the first place.

PR #1218, opened 2016-08-30
AKU-1080: Ensure DialogService clones widget models

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1080 to ensure that the DialogService clones widget models that are provided. This ensures that repeated invocations of the same dialog (in the failing case a dialog containing a form) will not use a model updated (for example with pubSubScope data) with the first invocation. A unit test has been added that replicates the failing use case to ensure that it is fixed.


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -474,8 +474,8 @@ define(["dojo/_base/declare",
             duration: payload.duration || 0,
             fullScreenMode: payload.fullScreenMode || false,
             fullScreenPadding: !isNaN(payload.fullScreenPadding) ? payload.fullScreenPadding : 10,
-            widgetsContent: payload.widgetsContent,
-            widgetsButtons: payload.widgetsButtons,
+            widgetsContent: lang.clone(payload.widgetsContent),
+            widgetsButtons: lang.clone(payload.widgetsButtons),
```
MartinDoyleUK: Just to double-check that the widgetsContent and widgetsButtons are either never undefined/null or do not cause a problem with lang.clone if they are?

draperd: This doesn't appear to cause a problem, in the unit test page we have widgetsContent but not widgetsButtons and that doesn't result in any errors

PR #1219, opened 2016-08-30
AKU-1076: Ensure filteringTopics do not need to be configured for AlfFilteredList

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1076 to ensure that it is not necessary to explicitly configure the filteringTopics array for an AlfFilteredList. It was not necessary to add an existing unit test - the existing tests failed when the filteringTopics configuration is removed from the current test page - updating the code resolved those failures.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/AlfFilteredList.js
+++ b/aikau/src/main/resources/alfresco/lists/AlfFilteredList.js
@@ -466,6 +466,8 @@ define(["dojo/_base/declare",
       setupFilteringTopics: function alfresco_lists_AlfFilteredList__setupFilteringTopics(filter) {
          if (filter && filter.config && filter.config.fieldId)
          {
+            // See AKU-1076 - ensure filteringTopics do not require explicit configuration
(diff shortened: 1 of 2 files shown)


aikau/src/main/resources/alfresco/lists/AlfFilteredList.js
```diff
@@ -466,6 +466,8 @@ define(["dojo/_base/declare",
       setupFilteringTopics: function alfresco_lists_AlfFilteredList__setupFilteringTopics(filter) {
          if (filter && filter.config && filter.config.fieldId)
          {
+            // See AKU-1076 - ensure filteringTopics do not require explicit configuration
+            !!!this.filteringTopics && (this.filteringTopics = []);
```
MartinDoyleUK: Three exclamation marks seems an error. Either one or two (presumably 1?).

Also, the fact that you had to surround the second expression with brackets suggests that it's not good to hide variable assignments in this way. Perhaps a simple `this.filteringTopics = this.filteringTopics || []` might be ok?

PR #1224, opened 2016-08-31
Expose simple value for FilePicker in "single selection" mode

AFaust:
Currently the FilePicker always exposes an array of (removed/added) values, even though the picker may be configured to only allow a single selection. For single-valued fields it is common (and expected) that the value is provided as a simple value and not wrapped inside a "dummy" array of a single element. This PR adds special handling for the case that multipleItemSelection is configured to false, yielding only the value of the itemKey without a wrapping array as the (added/removed) value of the field.

This PR also moves code that was specific to FilePicker from the BaseFormControl module (where it didn't belong) to the FilePicker itself.

Additionally it addresses a minor bug found during testing that could result in forms not being initialized correctly: calls to publishValue providing a Deferred may occur AFTER the widget has already been added to the DOM tree, resulting in the Deferred to never be resolved and thus the form to never complete its initialization.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/BaseFormControl.js
@@ -1917,64 +1917,9 @@ define(["dojo/_base/declare",
             var addedName = this.name + this.addedNameSuffix;
             var removedName = this.name + this.removedNameSuffix;
 
-            if (this.initialFileSelection)
-            {
-               var initialPick = this.initialFileSelection;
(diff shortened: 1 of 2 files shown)


draperd:
Thanks for raising this. I'll check out the branch and run the regression tests to verify nothing breaks (which I don't expect it will). I have to admit that I've implemented this picker in quite a hurry in order to progress with the FormsRuntimeService work so I'm very grateful to you for picking up these issues.

draperd:
Regressions test passing, code merged - thanks again!

Issue #1254, opened 2016-10-13
CoreXhr not depending on Core


HansDeBal:
Hi,

Just a remark/question: I see alfresco/core/CoreXhr does not require alfresco/core/Core (not in the define-parameters of CoreXhr). But since CoreXhr is using methods of Core (eg alfCleanFrameworkAttributes), shouldn't CoreXhr explicitly define a dependency on Core?

Thanks

draperd:
Yes, you're quite right... the reason why we've gotten away with this is because in almost all cases where a widget mixes in CoreXhr it is also mixing in Core. I'll get this changed. Thanks for raising this.

draperd:
I've created a pull request to address this - could you let me know if you're happy with the changes please?

PR #1258, opened 2016-10-18
Updates to CoreXhr module dependencies

draperd:
This PR addresses #1254 to update the dependencies of CoreXhr to include Core and to update those modules that were depending upon both CoreXhr and Core to just depend on CoreXhr.


HansDeBal:
I'm not an Aikau-specialist, but the changes look ok!

draperd:
@HansDeBal Thanks for taking a look at the change.

PR #1260, opened 2016-10-18
SHA-1807 - Create button "Create a link" in Copy dialog

rpopa:


aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -80,7 +80,17 @@ define(["dojo/_base/declare",
       moveAPI: "slingshot/doclib/action/move-to/node/",
 
       /**
-       * Sets up the service using the configuration provided. 
+       * URL to call to create a link to a document
+       *
+       * @instance
+       * @type {string}
+       * @default
+       */
```
draperd: When adding a new attribute or function to a module we add a "since" annotation so that the JSDoc shows when the capability was added, so here for example you'd want to add:

```
@since 1.0.91
```

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -124,10 +134,18 @@ define(["dojo/_base/declare",
...
+             createLinkPublishPayload = {
+                        nodes: nodes,
+                        documents: documents,
+                        responseScope: payload.alfResponseScope
+                     };
 
          var firstFileName = documents.length ? documents[0].fileName : null;
          var fileName = (nodes.length === 1 && firstFileName) ? firstFileName : this.message("services.ActionService.copyMoveTo.multipleFiles");
          this._actionHandle = this.alfSubscribe(responseTopic, lang.hitch(this, this.onCopyMoveLocationSelected, urlPrefix, payload.copy), true);
+         this._actionHandle = this.alfSubscribe(createLinkResponseTopic, lang.hitch(this, this.onCreateLinkLocationSelected), true);
```
draperd: It looks like you're replacing the previous _actionHandle set here, although looking at the code that was already there it looks like there was a bug... the `onCopyMoveLocationSelected` function is trying to remove `_actionCopyHandle` which has never been set. It looks like we need to do a bit of tidying up here.

I would suggest that we make `_actionHandle` become an array called `_actionHandles` and then we push both subscription handles into it. The ensure that the handles are unsubscribed on an action.

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -201,6 +234,42 @@ define(["dojo/_base/declare",
...
+
+       /**
+        * Handles the actual create link XHR call
+        *
+        * @instance
+        * @param {boolean} copy A boolean indicating if this is a copy action or not
+        * @param {object} payload The payload from confirmation of the action dialog
+        */
+      onCreateLinkLocationSelected: function alfresco_services_ActionService__onCreateLinkLocationSelected(payload) {
+          this.alfUnsubscribeSaveHandles([this._actionCopyHandle]);
```
draperd: This is where we'd pass `this._actionHandles` mentioned in the earlier comment... the `alfUnsubscribeSaveHandles` accepts an array, so we'd want to remove the square brackets.

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -124,10 +134,18 @@ define(["dojo/_base/declare",
...
+                        nodes: nodes,
+                        documents: documents,
+                        responseScope: payload.alfResponseScope
+                     };
 
          var firstFileName = documents.length ? documents[0].fileName : null;
          var fileName = (nodes.length === 1 && firstFileName) ? firstFileName : this.message("services.ActionService.copyMoveTo.multipleFiles");
          this._actionHandle = this.alfSubscribe(responseTopic, lang.hitch(this, this.onCopyMoveLocationSelected, urlPrefix, payload.copy), true);
+         this._actionHandle = this.alfSubscribe(createLinkResponseTopic, lang.hitch(this, this.onCreateLinkLocationSelected), true);
          this.alfPublish("ALF_CREATE_DIALOG_REQUEST", {
```
draperd: This isn't something I'd have expected you to have known... but since this function was originally written we've started using the `alfresco/core/topics` module for topic constants. So here we could use `topics.CREATE_DIALOG`. We'd also want to add a `@fires module:alfresco/core/topics#CREATE_DIALOG` statement to the JSDoc block...

...again, this isn't something I'd expect you to have known, but it would be useful if we can update modules as we go.

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -201,6 +234,42 @@ define(["dojo/_base/declare",
...
+        * @param {boolean} copy A boolean indicating if this is a copy action or not
+        * @param {object} payload The payload from confirmation of the action dialog
+        */
+      onCreateLinkLocationSelected: function alfresco_services_ActionService__onCreateLinkLocationSelected(payload) {
+          this.alfUnsubscribeSaveHandles([this._actionCopyHandle]);
+
+          // Get the locations to create link to and the documents to them...
+          if (payload.dialogContent)
+          {
+             when(payload.dialogContent, lang.hitch(this, function(content) {
```
draperd: Because this block is so similar to the one in `onCopyMoveLocationSelected` we could actually have "hitched" the same function on both subscriptions with an additional argument indicating the function to call if validation checks pass (e.g. `performCreateLinkAction` or `performAction`)

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -233,8 +302,33 @@ define(["dojo/_base/declare",
       },
 
       /**
+       * Makes an XHR call to create links for the selected documents to the location provided.
+       *
+       * @instance
+       * @param {array} nodeRefs An array of the nodes to create links for
+       * @param {array} location  The location to move or copy the documents to
+       */
```
draperd: As with other comments, we should add `@since 1.0.91` here

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -299,6 +393,44 @@ define(["dojo/_base/declare",
...
 
       /**
+       * Handles successful actions. When an attempt is made to create link to more than one node it is possible that
+       * only for some of the nodes it will be created successfully. If for all the nodes the link was created
+       * successfully then a simple notification is displayed indicating a successful action was completed, or partial
+       * completed.
+       *
+       * @instance
+       * @param {object} payload
+       * @fires module:alfresco/core/topics#DISPLAY_NOTIFICATION
```
draperd: '@since 1.0.91` again here

aikau/src/main/resources/alfresco/services/actions/CopyMoveService.js
```diff
@@ -317,6 +449,25 @@ define(["dojo/_base/declare",
             title: payload.requestConfig.copy ? this.message("copyMoveService.copy.failure.title") : this.message("copyMoveService.move.failure.title"),
             message: payload.requestConfig.copy ? this.message("copyMoveService.copy.failure") : this.message("copyMoveService.move.failure")
          });
+      },
+      /**
+       * Handles failed actions by displaying a notification indicating that the action was not successful.
+       *
+       * @instance
+       * @param {object} payload
+       * @fires module:alfresco/core/topics#DISPLAY_PROMPT
```
draperd: '@since 1.0.91` again here

draperd:
@rpopa I'm going to do the 1.0.91 release, so the `@since` tags will need to be 1.0.92.

draperd:
There's something else that I've realised - we need to have the create link button disabled by default so that newer releases of Aikau can be used against pre-5.2 releases of Share - we would need to have a configuration option in the CopyMoveService such as `supportLinkCreation` (or similar) with a default value of `false`. In the 5.2 code for Share we'd want to reconfigure the CopyMoveService so that `supportLinkCreation` would be true - this means that a non-functional "Create as link" button will not appear in 5.1

draperd:
Good stuff - thanks for making those changes. I've checked it out and tried the unit tests locally and it all seems good. Merging now.

PR #1265, opened 2016-10-24
AKU-1100: Check date is a Date before passing it to stamp.toISOString

davidcognite:


davidcognite:
Added ReleaseNote update as the ticket that introduced the bug wasn't listed.

draperd:
Thanks for finding and fixing this.

PR #1269, opened 2016-10-27
functionUtils missing internal flag for running period

AFaust:
While working on a Share-tier Log4J tool of https://github.com/AFaust/ootbee-support-tools I found that functionUtils would re-start a specific period over and over on each call to addRepeatingFunction. This was due to a flag not being set which is checked as the condition of the _startRepeatingTimeout call in addRepeatingFunction. In my case I needed a function to be called every 10s but since the period had been re-started a couple of times it was effectively triggered about every 2-4s due to ~5 active timeouts on the same period.
This PR adds the flag.


draperd:
Thanks for providing this... I've run the relevant tests and it seems good so I'll merge.

PR #1286, opened 2016-11-10
AKU-1119: Improve SiteService preset configuration options

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1119 to improve the configuration options for site presets in the create site dialog. It is now possible to replace the default options, add to the default options or remove from default and added options. Unit tests have been added to verify all the various configuration options.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/SiteService.js
+++ b/aikau/src/main/resources/alfresco/services/SiteService.js
@@ -80,8 +80,10 @@ define(["dojo/_base/declare",
       /**
        * This is an array of site preset options that will be displayed when creating a site (when the service
        * is not configured to be in [legacyMode]{@link module:alfresco/services/SiteService#legacyMode}). The
-       * default configuration just contains the "Collaboration Site", but this can be re-configured to contain
-       * additional site presets. Any additional presets must have values that map to presets in the Surf configuration.
+       * default configuration just contains the "Collaboration Site", but this can be overridden here. Any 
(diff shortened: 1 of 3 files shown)


davidcognite:
Added a commit to update the sitePreset method to make it clear it replaces the defaults.

aikau/src/main/resources/alfresco/services/SiteService.js
```diff
@@ -111,6 +142,20 @@ define(["dojo/_base/declare",
                { label: "create-site.dialog.type.collaboration", value: "site-dashboard" }
             ];
          }
+
+         if (this.additionalSitePresets && this.additionalSitePresets.length)
```
draperd: Well, this does a couple of things... firstly it is a crude sniff test to check that it *might* be an array, and if it is an array then using falsy logic it only proceeds with the concat operation if there actually is something concatenate.

davidcognite: it was the crude part that I didn't like; I tend to think either it's worth validating the input, or it's not, and if it's worth doing, then it's worth doing properly...

draperd: Even if I validated that it was an array, I'd still validate that there was something to concatenate. I'll add the additional verification if you insist.

davidcognite:
The approach is good, the solution will work for us. Happy for you to merge, unless you wanted to update the array validation.

PR #1305, opened 2016-11-23
AKU-1130: Improved create/edit site dialog customization options

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1130 to make it possible to customize the create and edit site dialog models in the SiteService through configuration. This capability is provide by a new WidgetsOverrideMixin module (intended for use in other widgets/services). Unit tests have been added to verify the capabilities.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/WidgetsOverrideMixin.js
+++ b/aikau/src/main/resources/alfresco/core/WidgetsOverrideMixin.js
@@ -1,0 +1,247 @@
+/**
+ * Copyright (C) 2005-2016 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * This mixin module provides a [function]{@link module:alfresco/core/WidgetsOverrideMixin#applyWidgetOverrides}
+ * for overriding the default widget model to be generated. This function provides a consistent way in which 
+ * new widgets can be added, removed, replaced and updated.
+ * 
+ * @module alfresco/core/WidgetsOverrideMixin
+ * @author Dave Draper
+ * @since 1.0.97
+ */
+define(["dojo/_base/declare",
+        "alfresco/core/ObjectTypeUtils",
+        "jquery"], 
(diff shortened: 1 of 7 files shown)


aikau/src/main/resources/alfresco/services/DialogService.js
```diff
@@ -573,7 +573,7 @@ define(["dojo/_base/declare",
             try
             {
                // Create a new pubSubScope just for this request (to allow multiple dialogs to behave independently)...
-               var pubSubScope = this.generateUuid();
+               var pubSubScope = payload.pubSubScope || this.generateUuid();
```
draperd: This code isn't actually changing anything that has really gone before (i.e. pubSubScope configuration in payload would have had no impact). This pubSubScope generation also was added before introduction of dialogId (for cleanup/re-use). I originally added this for testing purposes but then worked around it. I can remove it, however I can see the value here for customization of not just this dialog but others.

aikau/src/main/resources/alfresco/core/WidgetsOverrideMixin.js
```diff
@@ -0,0 +1,261 @@
...
+       * ]
+       *
+       * @instance
+       * @param {object[]} widgets The default widgets
+       * @param {object[]} overrides The overrides to apply
+       */
+      applyWidgetOverrides: function alfresco_core_WidgetsOverrideMixin__applyWidgetOverrides(widgets, overrides) {
+         var existingWidgetIndex;
+
+         if (ObjectTypeUtils.isArray(widgets) && ObjectTypeUtils.isArray(overrides))
```
davidcognite: Worth adding a log message in the else block rather than just silently failing if a object is provided?

aikau/src/main/resources/alfresco/services/SiteService.js
```diff
@@ -30,15 +30,16 @@ define(["dojo/_base/declare",
         "alfresco/core/NotificationUtils",
         "alfresco/core/ObjectTypeUtils",
         "alfresco/core/topics",
+        "alfresco/core/WidgetsOverrideMixin",
```
davidcognite: Great idea with the mixin. Have you got tickets to add this mixin to other widgets that include default widget config?

aikau/src/main/resources/alfresco/services/DialogService.js
draperd: What's life without a bit of danger ;)

aikau/src/main/resources/alfresco/core/WidgetsOverrideMixin.js
draperd: Yes, definitely... I'll push that update

aikau/src/main/resources/alfresco/services/SiteService.js
```diff
@@ -64,6 +65,34 @@ define(["dojo/_base/declare",
...
+       * (for example a custom preset). The use case is when a custom field needs to make an XHR request
+       * for data to be added (i.e. something not possible through 
+       * [autoSetConfig]{@link module:alfresco/forms/controls/BaseFormControl#autoSetConfig}).
+       * 
+       * @instance
+       * @type {string}
+       * @default
+       * @since 1.0.97
+       */
+      setCreateSiteDialogValueTopic: null,
```
draperd: This is a specific requirement of a customer.

davidcognite:
LGTM, with minor comments as noted

PR #1322, opened 2016-12-03
Demonstrate overhead of multiple widget config properties

AFaust:
This PR extends the performance test to showcase scaling issues with more complex widget configurations. All the properties being set are non-functional for the widget but current handling (most prominently the inherited _applyAttributes from Dojo/Diji _WidgetBase) adds significant overhead.


draperd:
Thanks Axel - this was one area I hadn't really looked at in depth yet from your original performance review.

PR #1323, opened 2016-12-03
Avoid _applyAttributes overhead by "hiding" complex Aikau widget configuration

AFaust:
Aikau widget configurations can contain an arbitrary number of configuration attributes. The way these are handled currently includes them in the "param" object which _applyAttributes in Dijit _WidgetBase processes. This is meant to apply any attributes that can be mapped to DOM or a widget-internal setXY operation, which does not apply for 99+% of Aikau attributes. All the lookup / looping over Aikau properties can introduce significant overhead for non-trivial widgets.

This PR adds extra handling to postMixInProperties and postCreate of Row, Cell and Property modules to avoid such overhead. For some reason it was not possible to simply override _applyAttributes to achieve the same.
It is not ideal that this is added in the "leaf" widget modules but unfortunately Aikau does not really use a common "base widget" module at this time. Also some widget modules like AlfMenuItem that extend non-trivial Dijit widgets may need to expose a sub-set of the configuration properties to _applyAttributes. Ideally either a new common "base widget" module or ChildProcessing could add logic to handle this without each widget having to do it itself or remember to include some mixin.

This PR is based on #1322 to be able to observe before/after performance measurements.

### DIFF
--- a/aikau/src/main/resources/alfresco/lists/views/layouts/Cell.js
+++ b/aikau/src/main/resources/alfresco/lists/views/layouts/Cell.js
@@ -93,6 +93,21 @@ define(["dojo/_base/declare",
             this.domNode.classList.add("alfresco-lists-views-layouts-Cell");
          }
       },
+      
+      /**
+       * @instance
+       * @since 1.0.NEXT
+       */
+      postMixInProperties : function alfresco_lists_views_layouts_Cell__postMixInProperties() {
+          // this widget (as all Aikau widgets) can be configured with countless config attributes
+          // _applyAttributes causes significant overhead since it processes all "as if" they can be mapped to DOM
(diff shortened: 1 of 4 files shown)


draperd:
Thanks for providing this... I'm going to merge this in, but I'm increasingly of the opinion that the branch isn't going to get merged as-is back into develop. I was hoping that I'd be able to make most of the changes fully backwards compatible, but it seems like I have to change the internal implementation in ways that I'm not comfortable with at the moment (like starting to depend on return values where previously there were none) which won't impact pure model usage, but could impact custom extensions.

I'm going to break all of these improvements into separate tasks and merge the "safe" ones directly into develop and then add the other ones into new widgets in the "aikau" package.

With regards to the lack of a common module to extend - I think that is a fair point and definitely something we can address.

PR #1330, opened 2016-12-05
AKU-1145, AKU-1147: Performance update pattern

draperd:
This PR addresses https://issues.alfresco.com/jira/browse/AKU-1145 and https://issues.alfresco.com/jira/browse/AKU-1147 to define a more performant approach for building widgets that removes the overhead of the _applyAttributes function and provides a new life-cycle function to be used to build the widget DOM natively (but still support extending widgets to provide templates). Although this doesn't tackle all Aikau widgets, it addresses the main widgets in the performance test page to try to demonstrate a path forwards that provides a rendering boost.

@AFaust I'd appreciate you taking a look at this.

### DIFF
--- a/aikau/src/main/resources/alfresco/core/BaseWidget.js
+++ b/aikau/src/main/resources/alfresco/core/BaseWidget.js
@@ -64,7 +64,7 @@ define(["dojo/_base/declare",
        */
       _applyAttributes: function alfresco_core_BaseWidget___applyAttributes() {
          var originalParams = this.params;
-         this.params = this.getParametersToApply();
+         this.params = this.getParametersToApply(originalParams);
          this.inherited(arguments);
          this.params = originalParams;
       },
@@ -80,7 +80,14 @@ define(["dojo/_base/declare",
        * @return {object} An object of key/value pair parameters to be applied. 
        */
       getParametersToApply: function alfresco_core_BaseWidget__getParametersToApply(/*jshint unused:false*/ originalParams) {
-         return null;
+         var params = null;
+         if (originalParams.style)
+         {
+            params = {
+               style: originalParams.style
+            };
+         }
+         return params;
       },
 
       /**
--- a/aikau/src/main/resources/alfresco/core/BaseWidget.js
+++ b/aikau/src/main/resources/alfresco/core/BaseWidget.js
@@ -1,0 +1,121 @@
+/**
+ * Copyright (C) 2005-2016 Alfresco Software Limited.
+ *
+ * This file is part of Alfresco
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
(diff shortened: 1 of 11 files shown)


draperd:
This is obviously only one part of a broader set of performance improvements initiatives.

draperd:
Two unit tests failing... investigating

draperd:
Unit tests now all passing, ready to merge when we're happy with this approach

aikau/src/main/resources/alfresco/core/BaseWidget.js
```diff
@@ -0,0 +1,120 @@
...
+       * to get an object of the parameters that should be applied to the widget. This should only
+       * be implemented as necessary. The more parameters applied the worse the performance at scale.
+       * 
+       * @instance
+       * @overridable
+       * @param {object} originalParams The original parameters provided to the widget.
+       * @return {object} An object of key/value pair parameters to be applied. 
+       */
+      getParametersToApply: function alfresco_core_BaseWidget__getParametersToApply(/*jshint unused:false*/ originalParams) {
+         return null;
```
draperd: I'm wondering if "style" should always be checked for here.

aikau/src/main/resources/alfresco/core/BaseWidget.js
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+/**
+ * Provides a base for all Aikau widgets to extend. It is not essential that a widget extends
+ * this module but it can provide significant performance benefits if hundreds or thousands
+ * of instances of the widget need to be rendered at a time. Any widget that does extend this
+ * base must either define a [templateString]{@link module:alfresco/core/BaseWidget#templateString}
+ * or override [createWidgetDom]{@link module:alfresco/core/BaseWidget#createWidgetDom}.
+ * 
+ * @module alfresco/core/BaseWidget
```
draperd: Should this actually be in the new "aikau" package?

aikau/src/main/resources/alfresco/core/BaseWidget.js
AFaust: I would say so. Although I also have some improvements for (extended) use of style in mind. When using repeating widgets (i.e. rows / cells / property renderers) it can result in extremely inefficient layout / paint cycles by the browser if we only use inline styling. At PRODYNA I had included dynamic stylesheet handling in our "high performance" base widget to address this.

draperd: OK... so I'm happy to incrementally nudge this forwards, but I'll make the change to add style. I'd be interested in getting your other updates though. It could be that the performance test page still isn't representative, but I've found (when applying all the various prototypes performance fixes) that their is actually very little time spent in paint cycles, but I'm happy to keep refining this once we have something stable and backwards compatible.

aikau/src/main/resources/alfresco/core/BaseWidget.js
```diff
@@ -0,0 +1,121 @@
...
+       * Extends the function inherited from dijit/_WidgetBase to only make use of the template
+       * rendering capabilities provided by dijit/_TemplatedMixin when a
+       * [templateString]{@link module:alfresco/core/BaseWidget#templateString} is provided. Otherwise
+       * the [createWidgetDom]{@link module:alfresco/core/BaseWidget#createWidgetDom} will be called
+       * to build the DOM for the widget.
+       * 
+       * @instance
+       */
+      buildRendering: function alfresco_core_BaseWidget__buildRendering() {
+         if (this.templateString)
```
AFaust: Before merging I'd like to refresh / check how _AttachMixin is tied into buildRendering cycle. I remember that for extensive DOM constructs (templated or otherwise) the lookup of any widgets / events to attach could get quite expensive and especially wastefull if there aren't any events / nodes to be attached at all or you are already handling it explicitly. I feel if we add the option for manual buildRendering we might as well include the same for attach handling - even if _TemplateMixin is used...

AFaust: Agreed with the long term aim and fine to go ahead with that in mind

PR #1345, opened 2016-12-26
Add value delimiter handling to Picker

AFaust:
Similar to the PR #1224 which added value delimiter handling (adding simple value support) to the FilePicker module this PR adds simple value support to Picker and derived modules, using an explicitly configured valueSeparator to determine if the control value should be stringified. Currently only the SitePicker module allows single-selection mode to be configured but in the future other this may be added to other picker modules.


draperd:
Thanks for raising this... I'll checkout the branch locally and run the regression tests.

draperd:
Merged. Thanks for another contribution.

PR #1346, opened 2016-12-26
Allow widgetsForControl configuration in SitePicker

AFaust:
Currently the widgetsForControl structure is initialised in the postMixInProperties function which prevents any customisation of the structure by simple JSON model configuration. The current initialisation can be moved to a simple prototype override, allowing configuration. This PR does just that.

### DIFF
--- a/aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
+++ b/aikau/src/main/resources/alfresco/forms/controls/SitePicker.js
@@ -77,122 +77,114 @@ define(["alfresco/forms/controls/Picker",
        */
       singleItemMode: true,
 
-      /**
-       * After config has been mixed in, setup the widget model
-       *
-       * @instance
-       */
-      postMixInProperties: function(){
-         this.widgetsForControl = [
-            {
-               name: "alfresco/layout/VerticalWidgets",
-               assignTo: "verticalWidgets",
-               config: {
-                  widgets: [
-                     {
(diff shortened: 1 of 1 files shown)


draperd:
The reason why the model was not defined as an attribute of the widget was because of the references to `this.singleItemMode` within the model. I think that this will result in that attribute always being false regardless of configuration overrides. Therefore the correct solution would be to take advantage of model processing (provided by alfresco/core/ObjectProcessingMixin) and use the `processInstanceTokens` processor and update the model to use a token that is substituted.

This requires a call to clone the model prior to use, a call to `processObject` and also requires that the "alfresco/pickers/SitePicker" also be updated.

As it is, I don't think I can merge this PR currently.

AFaust:
The base module Picker already uses ObjectProcessingMixin to process the widget. I already adapted the reference to itemKey to be a textual token ```{itemKey}```. I may have missed the reference to singleItemMode but strangely the changed variant worked without issues locally. I'll check and update the PR for any references I may have missed.

draperd:
OK, thanks.

AFaust:
Done - I added additional properties for the configuration of the relevant labels (select button, select dialog, remove button) and if they are not externally configured they are defaulted in postMixInProperties. Instance tokens in the widget model then refer to these properties.

draperd:
Thanks for making the changes, I'll run a regression test suite as soon as the one on your other PR completes.

draperd:
Tested and merged

PR #1360, opened 2017-01-26
Added the possibility to filter getSite requests in SiteService according to what the API provides

marsv024:
We are in need to filter getSite requests using the filters provided by the API. I added the filtering options to Site Service, however I'm unsure on how to implement a test for this. I could not find any previous tests for the getSites method, getUserSites method or getAdminSites method to use as a template. Please advise on how to proceed or if this commit is good enough.

### DIFF
--- a/aikau/src/main/resources/alfresco/services/SiteService.js
+++ b/aikau/src/main/resources/alfresco/services/SiteService.js
@@ -313,11 +313,52 @@ define(["dojo/_base/declare",
       },
 
       /**
+       * This returns all sites that the user has the permission to see. Eg. all public sites.
        *
        * @instance
        * @param {object} payload The details of the request
        */
       getSites: function alfresco_services_SiteService__getSites(payload) {
+        var url = AlfConstants.PROXY_URI + "api/sites";
+        var hasArguments = false;
+        //Add name filter if supplied
+        if (payload.nameFilter) {
+          if (hasArguments) {
+            url += "&";
+          } else {
+            url += "?";
+          }
+          url += "nf=" + payload.nameFilter;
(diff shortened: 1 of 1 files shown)


aikau/src/main/resources/alfresco/services/SiteService.js
```diff
@@ -313,11 +313,52 @@ define(["dojo/_base/declare",
...
        *
        * @instance
        * @param {object} payload The details of the request
        */
       getSites: function alfresco_services_SiteService__getSites(payload) {
+        var url = AlfConstants.PROXY_URI + "api/sites";
+        var hasArguments = false;
+        //Add name filter if supplied
+        if (payload.nameFilter) {
+          if (hasArguments) {
```
AFaust: I believe the code could be optimised before being merged by using the ```alfresco/util/urlUtils``` module to handle the URL construction. Specifically the ```addFilterQueryParameters``` operation can already deal with an arbitrary number of ```payload.dataFilter``` entries to be added to the URL.

aikau/src/main/resources/alfresco/services/SiteService.js
```diff
@@ -313,11 +313,52 @@ define(["dojo/_base/declare",
...
+            url += "&";
+          } else {
+            url += "?";
+          }
+          url += "roles=" + payload.roles;
+        }
          // TODO: Clean this up. Choose on or other as the Aikau standard.
          var alfResponseTopic = payload.alfResponseTopic || payload.responseTopic;
          var config = {
```
AFaust: Aren't you missing a change to set the URL inside the ```config``` object?

PR #1361, opened 2017-02-18
Make hard-coded topics for sort requests / field selection configurable

### LINKED ISSUE #1352: Hardcoded uses of topics.SORT_LIST despite AlfList allowing configuration of sortRequestTopic
The AlfList module allows configuration of the sortRequestTopic so it may differ from the default value inherited from the _AlfDocumentListTopicMixin, and the SortOrderToggle module even allows configuration of the subscriptionTopic and oNConfig/offConfig constructs (which include a publishTopic). Despite this, several uses in HeaderCell and AlfSortablePaginatedList are hard-coded to use the topics.SORT_LIST constant, which renders all of the configurability moot.

As I may have mentioned sometime in the past: I consider the use of any topics.XY constant problematic when it is used for anything other than setting the default value of a configuration property. If there are topics that really need to be internal then these should not be configurable in any module. If at least one module allows configuration of a topic that uses a constant as a default value, that contant should never be used for hard-coding subscription or publication calls.

AFaust:
Resolves #1352

This PR addresses the points I reported in #1352. I have update all relevant references to topics.SORT_LIST to use an instance-configurable property instead which is by default initialised to the value of the topic constant, thus keeping backwards compatibility. In one instance I also corrected an incorrect use of the topics.UPDATE_LIST_SORT_FIELD constant when a relevant instance-level property already existed. For the SortOrderToggle module I had to move the ```onConfig```/```offConfig``` default override into a new ```postMixInProperties``` to deal with the configurable ```selectionTopic``` so users of this module are not forced to override/configure those two properties just because they configured a different ```selectionTopic```.

### DIFF
--- a/aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
+++ b/aikau/src/main/resources/alfresco/documentlibrary/_AlfDocumentListTopicMixin.js
@@ -225,7 +225,7 @@ define(["dojo/_base/declare",
       sortRequestTopic: topics.SORT_LIST,
       
       /**
-       * @event
+       * @event sortFieldSelectionTopic
        * @instance
        * @type {string} 
        * @default [UPDATE_LIST_SORT_FIELD]{@link module:alfresco/core/topics#UPDATE_LIST_SORT_FIELD}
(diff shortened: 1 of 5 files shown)


davidcognite:
This looks good to me, a safe, backwards compatible improvement.

PR #1387, opened 2017-10-03
Add HTML encoding in AlfBreadcrumb to avoid XSS issues.

tpage-alfresco:


tpage-alfresco:
I have raised a ticket to add a unit test here: https://github.com/Alfresco/Aikau/issues/1388

Issue #1388, opened 2017-10-03
Add unit test for change to AlfBreadcrumb


tpage-alfresco:
In https://github.com/Alfresco/Aikau/pull/1387 we added a call to encodeHTML to prevent a XSS issue in the breadcrumb trail. We should add a unit test to ensure this change is not regressed.

Currently the unit test framework is not working locally (at least I can't see how to run them - see also https://github.com/Alfresco/Aikau/issues/1359 ) and so it's basically impossible for me to write unit tests.

tpage-alfresco:
We also need to add a unit test for https://github.com/Alfresco/Aikau/pull/1389.

tunaaksoy:
A unit test for #1390 has to be implemented as well.

PR #1389, opened 2017-10-04
Escape the label of AlfButton to prevent XSS issues.

tpage-alfresco:


tpage-alfresco:
I have updated the ticket to add a unit test here: #1388

PR #1390, opened 2017-10-06
Adding an XSS fix for AlfListView

wossabi:


tunaaksoy:
A unit test has to be implemented for this see [https://github.com/Alfresco/Aikau/pull/1388](1388)

PR #1409, opened 2018-12-04
MNT-20222 - removed encoding of root label for copyMoveTo action

astrachan:


mikeh:
I believe this fix is good, as this `Tree` component appears to only be instantiated with a static `rootLabel` string, e.g. `"filepicker.sites.tree.root.label"` and therefore is not susceptible to code injection after the proposed fix.

astrachan:
Thanks @alexbalmus / @mikeh, I've provided further investigation via MNT-20222 which hopefully now shows where the original/first encoding takes place.

alexbalmus:
Ok, thank Alex. I'll proceed with the merge.

PR #1414, opened 2020-07-03
Fix/apps 228

anechifor:
Added travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache:
 branches:
   only:
     - develop
-    - ^APPS\/.*
+    - APPS*
     - /support\/.*/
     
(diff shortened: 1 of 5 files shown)


cristina-sirbu:
Why do we run WhiteSource and Veracode?

.travis.yml
```diff
@@ -0,0 +1,58 @@
...
+  - openjdk11
+
+cache:
+  directories:
+    - $HOME/.m2
+
+branches:
+  only:
+    - develop
+    - fix/APPS-228
```
cristina-sirbu, line 14: Maybe we should have a regex here to run the build on the development branches as well? (APPS-*) ?

PR #1422, opened 2021-06-08
APPS-1026: Added selectProperty to Aikau Selector component

bogdanioan-gabor:
Added selectProperty to Aikau Selector component.


alexbalmus (approved):
LGTM

PR #1434, opened 2022-04-11
[MNT-22946] Upgrade spring - RCE databind jdk9

jottley:
-Upgrade transative dependencies


vrmoreira (approved):
Running `mvn dependency:tree -Dincludes=org.springframework\*:` shows only spring version `5.3.18` (the one that we want)

So, LTGM!

tiagosalvado10 (approved):

